PR #7, opened 2025-10-07
 feat: add creator tracking and member management operations

### LINKED ISSUE #1: Message and Discussion models
Create the Message and Discussion models and test them.
Integrate them with Firebase

AntoineBastide47:
Add creatorId field to Discussion model to track the discussion owner, enabling owner-only operations such as deleting discussions and removing participants. Implement methods for removing users and admins from discussions with appropriate permission checks. Additionally, optimise the sendMessage operation to use batch writes for atomic updates to discussion messages and participant previews.

### DIFF
--- a/app/src/androidTest/java/com/github/meeplemeet/model/integration/FirestoreDiscussionTests.kt
+++ b/app/src/androidTest/java/com/github/meeplemeet/model/integration/FirestoreDiscussionTests.kt
@@ -184,12 +184,12 @@ class FirestoreDiscussionTests : FirestoreTests() {
   fun canChangeDiscussionDescription() = runTest {
     val (acc, discussion) = viewModels[0].createDiscussion("Test", "", account1)
     account1 = acc
-    val newDescription = "A non empty discussion"
+    val newDescription = "A non empty description"
     val updated = viewModels[1].setDiscussionDescription(discussion, account1, newDescription)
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/meeplemeet/model/systems/FirestoreRepository.kt
Atomium90, line 1: Looks great!
Just an idea (if feasable): maybe you could consider a private helper that handles property updates, so that all these `setProperty`, `addX` `removeY` methods just delegate to it?
Other than that possible refactor, everything looks good to me!

app/src/main/java/com/github/meeplemeet/model/viewmodels/FirestoreViewModel.kt
Atomium90, line 1: Here everything looks good. 
I would just make the same comment as for `FirestoreRepository.kt`

app/src/androidTest/java/com/github/meeplemeet/model/integration/FirestoreDiscussionTests.kt
Atomium90, line 1: Very good. These tests nicely cover the new functions.

Atomium90 (approved):
Looks clean and well done! ✅
Nothing major to add — just small optional improvements in the comments. Good to go from my side!

PR #21, opened 2025-10-09
Fix the soft errors outputed when calling sonar

AntoineBastide47:


Atomium90 (approved):
Nothing to say. CI change only.

PR #40, opened 2025-10-11
Try a new version of the CI

AntoineBastide47:
This PR brings changes to the CI to make it faster by running the unit and android tests on seperate jobs and now explicitly fails if the line coverage is less than 80% by running a new job after the coverage is created and sent to sonar qube.


Atomium90 (approved):
CI seems good. All checks pass.

PR #42, opened 2025-10-12
fix(tests): add email field to account setup

NilSamaha:
Fixes the CI failure since account now takes an email.


AntoineBastide47 (approved):
All good

PR #44, opened 2025-10-13
Add user handles

### LINKED ISSUE #30: Create the account creation with unique handles system

AntoineBastide47:
This PR implements a user handle system that allows users to create and manage unique handles (usernames) for their accounts, similar to Twitter/Instagram handles. Key features include:

  - Handle CRUD operations: Create, read, update, and delete user handles with proper validation
  - Handle uniqueness enforcement: Ensures handles are unique across the platform using Firestore transactions
  - Handle validation: Enforces format requirements (alphanumeric, underscores, hyphens) and size limits
  - New repository layer: Introduces FirestoreHandlesRepository for handle-specific operations
  - Authentication integration: Updated AuthRepository (renamed from AuthRepoFirebase) to work with the new handle system
  - Comprehensive testing: Added extensive integration tests for handle operations (524 new test lines)
  - CI/CD improvements: Fixed unit test coverage reporting and test infrastructure

  Technical Details

  New Components:
  - FirestoreHandlesRepository.kt - Repository for handle CRUD operations with transaction support
  - FirestoreHandlesViewModel.kt - ViewModel managing handle state and validation
  - Account.handle field - Added handle property to Account data structure
  - Handle validation logic with format and size constraints

  Architecture Changes:
  - Renamed AuthRepoFirebase → AuthRepository for consistency
  - Moved Google Sign-In helper to repositories package
  - Reorganized test structure (androidtest → integration package)
  - Enhanced exception handling with HandleAlreadyTakenException

  CI/CD Enhancements:
  - Fixed unit test artifact upload/download
  - Switched coverage metric to instruction coverage
  - Added lint report downloading
  - Updated dependency resolution for security vulnerabilities

  ---
  Generated with https://claude.com/claude-code


dtbcds:
This is a great addition, but I do have some questions:

- The current implementation uses Firestore transactions to ensure handle uniqueness, which is good. Is there a rate limit to avoid race conditions or excessive handle checks?

- Are there any validation rules for allowed handle formats (length, allowed characters)? Consider enforcing these server-side even if there will be checks client-side with the UI.

- When a user changes their handle, is the change propagated everywhere the handle is used? Should the places in the code or UI that still rely on UID be changed to all use handles?

AntoineBastide47:
> This is a great addition, but I do have some questions:
> 
> * The current implementation uses Firestore transactions to ensure handle uniqueness, which is good. Is there a rate limit to avoid race conditions or excessive handle checks?
> * Are there any validation rules for allowed handle formats (length, allowed characters)? Consider enforcing these server-side even if there will be checks client-side with the UI.
> * When a user changes their handle, is the change propagated everywhere the handle is used? Should the places in the code or UI that still rely on UID be changed to all use handles?

1) There can't be race conditions in firebase as everything is sequential and atomic, the transactions also garantee this.
2) Handles now only allow letters and digis.
3) This is done thanks to the account variable in the view model.

dtbcds (approved):
Looks good to me after the changes.

PR #46, opened 2025-10-14
fix(test): add missing currentUser parameter to DiscussionSettingScreen

NilSamaha:
- include currentUser (Account) as a required parameter for DiscussionSettingScreen


AntoineBastide47 (approved):
Thanks for the test bug fix.

PR #109, opened 2025-10-29
Feat/poll backend

### LINKED ISSUE #61: Integrate the polls with the backend

EnesKiper:
This pull request adds support for polls within discussions, allowing users to create polls, vote on options, and remove their votes. It introduces a new `Poll` data structure, extends the `Message` model to optionally include poll data, and implements all necessary repository and view model methods for poll management.

**Poll Feature Implementation:**

* Introduced a new `Poll` data class in `Poll.kt`, which includes poll question, options, votes, and whether multiple votes are allowed, along with utility methods for vote management and statistics.
* Updated the `Message` data class in `Message.kt` to optionally include a `poll` property, enabling messages to carry poll data.
* Added poll creation, voting, and vote removal methods in `FirestoreRepository.kt`, supporting poll lifecycle operations and synchronizing poll state with Firestore.
* Added corresponding poll management methods in `FirestoreViewModel.kt` to provide poll functionality to the UI layer, including input validation and repository delegation.

**Message Handling Enhancements:**

* Refactored message sending logic in `FirestoreRepository.kt` to handle both regular and poll messages, updating discussion previews accordingly and providing an overload that accepts a pre-constructed `Message` object. 

**Dependency Updates:**

* Imported the new `Poll` class in `FirestoreRepository.kt` to support poll-related features.

(done with copilot)


app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
```diff
@@ -451,4 +493,146 @@ class FirestoreRepository(db: FirebaseFirestore = FirebaseProvider.db) {
...
+   * @param optionIndex The index of the option to vote for.
+   */
+  suspend fun voteOnPoll(
+      discussionId: String,
+      pollMessageTimestamp: Timestamp,
+      userId: String,
+      optionIndex: Int
+  ) {
+    // Fetch discussion to get current state
+    val discussion = getDiscussion(discussionId)
```
AntoineBastide47: You can pass the whole discussion object instead, this will be faster than doing a get request (removes 10-500ms per vote).

app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
```diff
@@ -451,4 +493,146 @@ class FirestoreRepository(db: FirebaseFirestore = FirebaseProvider.db) {
...
+
+    // Find the message by timestamp (stable identifier)
+    val messageIndex = discussion.messages.indexOfFirst { it.createdAt == pollMessageTimestamp }
```
AntoineBastide47: This could be problematique, consider passing the message containign the poll instead, or at least it's index.

app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
```diff
@@ -451,4 +493,146 @@ class FirestoreRepository(db: FirebaseFirestore = FirebaseProvider.db) {
...
+      // Single vote mode: replace with single option
+      updatedVotes[userId] = listOf(optionIndex)
+    }
+
+    val updatedPoll = poll.copy(votes = updatedVotes)
+    val updatedMessage = message.copy(poll = updatedPoll)
+    val updatedMessages = discussion.messages.toMutableList()
+    updatedMessages[messageIndex] = updatedMessage
+
+    discussions.document(discussionId).update(Discussion::messages.name, updatedMessages).await()
```
AntoineBastide47: When you add the message or it's index as a param, you can directly update the message instead to reduce latency as updating the whole discussion does not scale. It should be (but check again just in case):
```kotlin
.update("${Discussion::messages.name}.$messageIndex", updatedMessages)
```

app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
```diff
@@ -451,4 +493,146 @@ class FirestoreRepository(db: FirebaseFirestore = FirebaseProvider.db) {
...
+   * Remove a user's vote for a specific poll option. Called when user clicks an option they
+   * previously selected to deselect it.
+   *
+   * @param discussionId The ID of the discussion containing the poll.
+   * @param pollMessageTimestamp The timestamp of the message containing the poll (used as
+   *   identifier).
+   * @param userId The user's UID.
+   * @param optionIndex The specific option index to remove.
+   */
+  suspend fun removeVoteFromPoll(
```
AntoineBastide47, line 595: Same remarks as for `voteOnPoll`

app/src/main/java/com/github/meeplemeet/model/structures/Poll.kt
```diff
@@ -0,0 +1,78 @@
...
+ * @property question The poll question or title.
+ * @property options List of poll options that users can vote for.
+ * @property votes Map of userId to the list of option indices they voted for.
+ * @property allowMultipleVotes Whether users can vote for multiple options.
+ */
+@Serializable
+data class Poll(
+    val question: String = "",
+    val options: List<String> = emptyList(),
+    val votes: Map<String, List<Int>> = emptyMap(), // Changed to List<Int>
```
AntoineBastide47: Just an idea, but changing options to: List<String, List<String>> might be better ? So each option also holds which accounts voted for it and you get the number of votes by using `.length`

app/src/main/java/com/github/meeplemeet/model/viewmodels/FirestoreViewModel.kt
AntoineBastide47, line 1: The comments to the repository also apply here

AntoineBastide47 (changes requested):
This PR is overall a very good addition. I just have some nitpicks that I would like changed. And don't forget to add some integration tests to make sure it also works on the firebase side.

app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
EnesKiper: ok, thought i might be too much to pass the whole discussion

app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
EnesKiper: you're right, i thought it would be too much to pass the whole message in the function, i will change it to pass the whole message

app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
EnesKiper: done

app/src/main/java/com/github/meeplemeet/model/structures/Poll.kt
EnesKiper: I don't think so, the most frequent actions are checking which options the user already voted for when voting, changing, or removing his vote, then doing the changes of changing a vote, removing, or adding a vote.

Checking how many votes an option has is just used by the UI for rendering. This is why i added a function getVoteCountsByOption()

app/src/main/java/com/github/meeplemeet/model/repositories/FirestoreRepository.kt
EnesKiper: i tried doing this, it looks like it is impossible to only update one element in an array in firestore.
Don't really know how to implement it

app/src/main/java/com/github/meeplemeet/model/viewmodels/FirestoreViewModel.kt
EnesKiper: done

AntoineBastide47 (approved):
Thanks for the changes

PR #124, opened 2025-11-03
Feat/poll UI

### LINKED ISSUE #65: Add the poll to the discussion screen

dtbcds:
# Add interactive polls to discussions

- Users can now create polls (2–7 options, single or multiple choice) straight from the attachment menu
- Live results are shown as percentage bars that update instantly when anyone votes
- Votes can be changed or revoked; single-choice polls auto-remove the previous selection
- Full test coverage for dialog UI, vote buttons and result display

poll-ezgif com-resize


AlexandroHattoum (changes requested):
### Overall
- Your code is clean, it's easy to follow and accomplishes the required feature. Good job here, nothing else to say.
- Documentation wise, you need Kdoc comments for your composables, it'll make revisiting your code in the future a breeze.
- Testing wise, you *should* have enough coverage according to Jacoco but we have to wait on Sonar for the real judgement. Additionally, the tests passed locally for me (and I'd guess you too) but the CI failed at the time of writing this. 
- UI wise, I think the poll bubble should be padded right/left like the messages, and the popup upon clicking the `attachement` icon should be a bit lower, and perhaps with a border. Experiment a bit with a redesign of that popup. 
A last nitpicky comment is to add a check icon to the selected option like the Figma design.

AlexandroHattoum (approved):
I've reviewed your modifications, thank you for the UI changes! I can also see you've got Kdocs now for your composables, which is great.
To me this is ready for merge, just check on the CI if it errors somewhere.

PR #132, opened 2025-11-07
Map screen integration with pins & bottom display

### LINKED ISSUE #120: Build Map Screen UI and Interactions
Implement the frontend map screen

- Create a screen where the map is visible
- Make it accessible from the bottom navbar
- Display pins with correct styling and data
- Add on-click callbacks for pins
- Add a button in the session view to open the map and show session location
- Add a button to create a new shop from the map screen

Atomium90:
### MapViewModel
- Add `MapViewModel` to manage geoQuery lifecycle, pin selection, and preview resolution
- Support type filtering, radius updates, and camera-based center repositioning
- Handle error messages and UI state updates

### MapScreen
- Create `MapScreen` composable with GoogleMap integration and dynamic marker rendering
- Display bottom sheet previews for Shop, Space, and Session pins
- FAB visible for shopOwner and spaceRenter accounts, with callback support
- Snackbar integration for error feedback

### Tests
- `MapViewModelTest`: covers lifecycle, selection, error handling, and state clearing
- `MapScreenTest`: validates UI rendering, FAB behavior, preview flow, and user interactions

### Misc
- Reorganized misplaced files for better structure
- Fix minor bugs in location search fields accross the app (shared component)
- Introduce `OnMapsSdkInitializedCallback` interface to stabilize Maps SDK initialization in test

_This PR was reformulated with Copilot to improve readability._


app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+      radiusKm: Double,
+      includeTypes: Set<PinType> = setOf(PinType.SHOP, PinType.SPACE)
+  ) {
+    stopGeoQuery()
+
+    val geoFirestore = GeoFirestore(geoPinCollection)
+    val query = geoFirestore.queryAtLocation(GeoPoint(center.latitude, center.longitude), radiusKm)
+
+    val listener =
+        object : GeoQueryEventListener {
```
AntoineBastide47, line 84: Add more docs for this part as it is an external API and most of us won't know what each function explicily does or when it is called.

app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
```diff
@@ -0,0 +1,161 @@
...
+
+  fun updateRadius(newRadiusKm: Double) {
+    geoQuery?.setRadius(newRadiusKm)
+  }
+
+  fun updateCenterAndRadius(center: Location, radiusKm: Double) {
+    geoQuery?.setLocation(GeoPoint(center.latitude, center.longitude), radiusKm)
+  }
+
+  fun selectPin(pin: GeoPinWithLocation) {
```
AntoineBastide47, line 217: When loading the GeoPins can't you also load there previews ? As they aren't very data heavy, you could load them after in the background so that clicking on a pin has little to no delay.

AntoineBastide47 (changes requested):
Overall, a very solid PR but some docs and optimizations should be added.

app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
Atomium90: Good idea, will do it

app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
Atomium90: The preview themselve are not heavy I agree, but to make a preview I need to load the whole data (shop, session...) So if for every pins I load the preview too, it will slow down everything. But I totally agree that there is optimization to be done, but it will involve architecural change. So it's a later task if youre ok with it.

AntoineBastide47: Yes, but shops and sessions aren't data heavy, so you can just delegate the loading to one background thread.

AntoineBastide47: TO DO in the very near future

AntoineBastide47 (approved):
Thanks for the docs

PR #139, opened 2025-11-11
Feat(ui): Implemented the SpaceRenterScreen

### LINKED ISSUE #98: Create the first draft of the space renter view screen

AlexandroHattoum:
This PR introduces the **SpaceRenterScreen** to the app, allowing users to view detailed information about a space renter. This includes the following:
- Contact Information: email, phone, website/informative link and their location.
- Availability, a 7 day schedule showing the availability of rentable spaces.
- Available rentable spaces, showing the number of seats available and their price.

### Additional Details
- The screen is tested and locally (with jacoco) achieves 80%+ coverage.
- Documentation is available for all of the UI components.

### Known Limitations
- The `edit` button does not navigate to the `EditSpaceRenterScreen` as it is not yet merged into main.
studio64_P4griW6uQN

studio64_GUnNPG0rkZ

https://i.imgur.com/oWkYqIF.gif

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
@@ -1,0 +1,287 @@
+// AI was used to help comment this screen
+package com.github.meeplemeet.ui.space_renter
+
+import android.widget.Toast
+import androidx.compose.foundation.layout.*
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.*
+import androidx.compose.material3.*
+import androidx.compose.runtime.*
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.platform.ClipboardManager
+import androidx.compose.ui.platform.LocalClipboardManager
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextIndent
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.viewmodel.compose.viewModel
+import com.github.meeplemeet.model.auth.Account
+import com.github.meeplemeet.model.shops.OpeningHours
+import com.github.meeplemeet.model.space_renter.SpaceRenter
+import com.github.meeplemeet.model.space_renter.SpaceRenterViewModel
+import com.github.meeplemeet.ui.components.SpacesList
+import com.github.meeplemeet.ui.components.TopBarWithDivider
+import com.github.meeplemeet.ui.theme.AppColors
+import java.text.DateFormatSymbols
+import java.util.Calendar
+
+/** Object containing test tags used in the Space Renter screen UI for UI testing purposes. */
+object SpaceRenterTestTags {
+  // Contact section tags
+  const val SPACE_RENTER_PHONE_TEXT = "SPACE_RENTER_TEXT"
+  const val SPACE_RENTER_PHONE_BUTTON = "SPACE_RENTER_PHONE_BUTTON"
+  const val SPACE_RENTER_EMAIL_TEXT = "SPACE_RENTER_EMAIL_TEXT"
+  const val SPACE_RENTER_EMAIL_BUTTON = "SPACE_RENTER_EMAIL_BUTTON"
+  const val SPACE_RENTER_ADDRESS_TEXT = "SPACE_RENTER_ADDRESS_TEXT"
+  const val SPACE_RENTER_ADDRESS_BUTTON = "SPACE_RENTER_ADDRESS_BUTTON"
+  const val SPACE_RENTER_WEBSITE_TEXT = "SPACE_RENTER_WEBSITE_TEXT"
+  const val SPACE_RENTER_WEBSITE_BUTTON = "SPACE_RENTER_WEBSITE_BUTTON"
+  const val SPACE_RENTER_EDIT_BUTTON = "EDIT_SPACE_BUTTON"
+
+  // Availability section tags
+  const val SPACE_RENTER_DAY_PREFIX = "SPACE_RENTER_DAY_"
+}
+
+/**
+ * Composable that displays the Space Renter screen, including the top bar and Space Renter details.
+ *
+ * @param spaceId The ID of the Space Renter to display.
+ * @param account The current user account.
+ * @param viewModel The ViewModel providing space renter data.
+ * @param onBack Callback invoked when the back button is pressed.
+ * @param onEdit Callback invoked when the edit button is pressed.
+ */
+@Composable
+fun SpaceRenterScreen(
+    spaceId: String,
+    account: Account,
+    viewModel: SpaceRenterViewModel = viewModel(),
+    onBack: () -> Unit = {},
+    onEdit: (SpaceRenter?) -> Unit = {},
+) {
+  // Collect the current space renter state from the ViewModel
+  val spaceState by viewModel.spaceRenter.collectAsStateWithLifecycle()
+  // Trigger loading of space renter data when spaceId changes
+  LaunchedEffect(spaceId) { viewModel.getSpaceRenter(spaceId) }
+
+  Scaffold(
+      topBar = {
+        TopBarWithDivider(
+            text = spaceState?.name ?: "Space Renter",
+            onReturn = { onBack() },
+            trailingIcons = {
+              // Show edit button only if current account is the space renter owner
+              if (account == (spaceState?.owner ?: false)) {
+                IconButton(
+                    onClick = { onEdit(spaceState) },
+                    modifier = Modifier.testTag(SpaceRenterTestTags.SPACE_RENTER_EDIT_BUTTON)) {
+                      Icon(Icons.Default.Edit, contentDescription = "Edit")
+                    }
+              }
+            })
+      }) { innerPadding ->
+        // Show space renter details if loaded, otherwise show a loading indicator
+        spaceState?.let { space ->
+          SpaceRenterDetails(
+              spaceRenter = space,
+              modifier = Modifier.padding(innerPadding).padding(16.dp).fillMaxSize())
+        }
+            ?: Box(modifier = Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
+              CircularProgressIndicator()
+            }
+      }
+}
+
+/**
+ * Composable that displays detailed information about a space renter, including contact info,
+ * availability, and game list.
+ *
+ * @param spaceRenter The space renter data to display.
+ * @param modifier Modifier to be applied to the layout.
+ */
+@Composable
+fun SpaceRenterDetails(spaceRenter: SpaceRenter, modifier: Modifier = Modifier) {
+  Column(modifier = modifier, verticalArrangement = Arrangement.spacedBy(24.dp)) {
+    ContactSection(spaceRenter)
+    HorizontalDivider(modifier = Modifier.fillMaxWidth().padding(horizontal = 100.dp))
+    AvailabilitySection(spaceRenter.openingHours)
+    HorizontalDivider(modifier = Modifier.fillMaxWidth().padding(horizontal = 100.dp))
+    Column(
+        verticalArrangement = Arrangement.spacedBy(8.dp),
+        modifier = Modifier.fillMaxWidth().padding(horizontal = 30.dp)) {
+          Text(
+              "Available spaces",
+              style = MaterialTheme.typography.titleLarge,
+              fontWeight = FontWeight.SemiBold)
+
+          SpacesList(
+              spaces = spaceRenter.spaces,
+              modifier = Modifier.fillMaxWidth(),
+              onChange = { _, _ -> },
+              onDelete = {},
+              isEditing = false,
+          )
+        }
+  }
+}
+
+// -------------------- CONTACT SECTION --------------------
+
+/**
+ * Composable that displays the contact information section of the space renter.
+ *
+ * @param spaceRenter The space renter whose contact information is displayed.
+ */
+@Composable
+fun ContactSection(spaceRenter: SpaceRenter) {
+  Column(
+      verticalArrangement = Arrangement.spacedBy(8.dp),
+      modifier = Modifier.fillMaxWidth().padding(horizontal = 25.dp)) {
+        Text(
+            "Contact",
+            style = MaterialTheme.typography.titleLarge,
+            fontWeight = FontWeight.SemiBold)
+
+        // Display phone contact row
+        ContactRow(
+            Icons.Default.Phone,
+            "- Phone: ${spaceRenter.phone}",
+            SpaceRenterTestTags.SPACE_RENTER_PHONE_TEXT,
+            SpaceRenterTestTags.SPACE_RENTER_PHONE_BUTTON)
+
+        // Display email contact row
+        ContactRow(
+            Icons.Default.Email,
+            "- Email: ${spaceRenter.email}",
+            SpaceRenterTestTags.SPACE_RENTER_EMAIL_TEXT,
+            SpaceRenterTestTags.SPACE_RENTER_EMAIL_BUTTON)
+
+        // Display address contact row
+        ContactRow(
+            Icons.Default.Place,
+            "- Address: ${spaceRenter.address.name}",
+            SpaceRenterTestTags.SPACE_RENTER_ADDRESS_TEXT,
+            SpaceRenterTestTags.SPACE_RENTER_ADDRESS_BUTTON)
+
+        // Display website contact row
+        ContactRow(
+            Icons.Default.Language,
+            "- Website: ${spaceRenter.website}",
+            SpaceRenterTestTags.SPACE_RENTER_WEBSITE_TEXT,
+            SpaceRenterTestTags.SPACE_RENTER_WEBSITE_BUTTON)
+      }
+}
+
+/**
+ * Composable that displays a single row of contact information with an icon, text, and a button to
+ * copy the text to the clipboard.
+ *
+ * @param icon The icon to display for the contact method.
+ * @param text The contact text to display and copy.
+ * @param textTag The test tag for the text element.
+ * @param buttonTag The test tag for the copy button.
+ */
+@Composable
+fun ContactRow(icon: ImageVector, text: String, textTag: String, buttonTag: String) {
+  val clipboardManager: ClipboardManager = LocalClipboardManager.current
+  val context = LocalContext.current
+  Row(
+      verticalAlignment = Alignment.CenterVertically,
+      horizontalArrangement = Arrangement.spacedBy(8.dp),
+      modifier = Modifier.fillMaxWidth().padding(horizontal = 4.dp)) {
+        Text(
+            text,
+            style = LocalTextStyle.current.copy(textIndent = TextIndent(restLine = 8.sp)),
+            modifier = Modifier.weight(1f).testTag(textTag))
+        IconButton(
+            onClick = {
+              // Copy the contact text to the clipboard and show a toast confirmation
+              clipboardManager.setText(AnnotatedString(text))
+              Toast.makeText(context, "Copied to clipboard", Toast.LENGTH_SHORT).show()
+            },
+            content = { Icon(icon, contentDescription = null, tint = AppColors.neutral) },
+            modifier = Modifier.size(24.dp).testTag(buttonTag))
+      }
+}
+
+// -------------------- AVAILABILITY SECTION --------------------
+
+/**
+ * Composable that displays the space renter's opening hours for each day of the week.
+ *
+ * @param openingHours List of OpeningHours representing the space renter's weekly schedule.
+ */
+@Composable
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+  Column(modifier = modifier, verticalArrangement = Arrangement.spacedBy(24.dp)) {
+    ContactSection(spaceRenter)
+    HorizontalDivider(modifier = Modifier.fillMaxWidth().padding(horizontal = 100.dp))
+    AvailabilitySection(spaceRenter.openingHours)
+    HorizontalDivider(modifier = Modifier.fillMaxWidth().padding(horizontal = 100.dp))
+    Column(
+        verticalArrangement = Arrangement.spacedBy(8.dp),
+        modifier = Modifier.fillMaxWidth().padding(horizontal = 30.dp)) {
+          Text(
+              "Available spaces",
```
MarcoLourenco55: (nitpick) Consider using another text to describe it, like "Provided spaces", to avoid repetition of "available" with the previous section

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+
+// -------------------- AVAILABILITY SECTION --------------------
+
+/**
+ * Composable that displays the space renter's opening hours for each day of the week.
+ *
+ * @param openingHours List of OpeningHours representing the space renter's weekly schedule.
+ */
+@Composable
+fun AvailabilitySection(openingHours: List<OpeningHours>) {
```
MarcoLourenco55, line 235: Consider making this screen consistent with the others when showing opening hours. We used the am/pm format in other screens, not hh:mm

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+          if (entry.hours.isEmpty()) {
+            // No opening hours means the space renter is closed on this day
+            Row(
+                modifier =
+                    Modifier.fillMaxWidth()
+                        .testTag("${SpaceRenterTestTags.SPACE_RENTER_DAY_PREFIX}${entry.day}"),
+                horizontalArrangement = Arrangement.SpaceBetween) {
+                  Text(
+                      dayName,
+                      fontWeight = if (isToday) FontWeight.Bold else FontWeight.Normal,
```
MarcoLourenco55: Even if not consistent with the "ShopScreen" from last week, I think it would be more readable to put the actual day in focus or neutral color. What do you think ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+  LaunchedEffect(spaceId) { viewModel.getSpaceRenter(spaceId) }
+
+  Scaffold(
+      topBar = {
+        TopBarWithDivider(
+            text = spaceState?.name ?: "Space Renter",
+            onReturn = { onBack() },
+            trailingIcons = {
+              // Show edit button only if current account is the space renter owner
+              if (account == (spaceState?.owner ?: false)) {
```
MarcoLourenco55: Won't this like create a bug when comparing an "account" to a boolean (false), due to the parenthesis ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+import com.github.meeplemeet.ui.components.SpacesList
+import com.github.meeplemeet.ui.components.TopBarWithDivider
+import com.github.meeplemeet.ui.theme.AppColors
+import java.text.DateFormatSymbols
+import java.util.Calendar
+
+/** Object containing test tags used in the Space Renter screen UI for UI testing purposes. */
+object SpaceRenterTestTags {
+  // Contact section tags
+  const val SPACE_RENTER_PHONE_TEXT = "SPACE_RENTER_TEXT"
```
MarcoLourenco55: Shouldn't the text be "SPACE_RENTER_PHONE_TEXT" instead ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+                          modifier =
+                              Modifier.weight(1f)
+                                  .testTag(
+                                      "${SpaceRenterTestTags.SPACE_RENTER_DAY_PREFIX}${entry.day}"))
+                    } else {
+                      // Empty space for subsequent interval rows to align with day name column
+                      Text("", modifier = Modifier.weight(1f))
+                    }
+                    // Format the time interval or show "Closed" if times are null
+                    val timeText = if (start != null && end != null) "$start - $end" else "Closed"
```
MarcoLourenco55: Consider testing the opening hours first, to avoid displaying wrong information. Imagine for example the case where a time slot has the starting point (11pm) after the end point (5am).
Also, consider the case where the start == end, and overlaping opening hours

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+                      // Show the day name only on the first interval row
+                      Text(
+                          dayName,
+                          fontWeight = if (isToday) FontWeight.Bold else FontWeight.Normal,
+                          modifier =
+                              Modifier.weight(1f)
+                                  .testTag(
+                                      "${SpaceRenterTestTags.SPACE_RENTER_DAY_PREFIX}${entry.day}"))
+                    } else {
+                      // Empty space for subsequent interval rows to align with day name column
```
MarcoLourenco55: Instead of using an if-condition and then inputting and artificial space to compensate, wouldn't it be better to use a combination of Row and Columns directly such that they do it automatically ? Or at least use Spacers instead of empty texts. What do you think ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Consider removing repetition of code for the bold text that could be performed once in advance

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55, line 235: Consider refactoring this composable or use helper functions to reduce it's cognitive complexity, specially for the nested if-conditions.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+    spaceId: String,
+    account: Account,
+    viewModel: SpaceRenterViewModel = viewModel(),
+    onBack: () -> Unit = {},
+    onEdit: (SpaceRenter?) -> Unit = {},
+) {
+  // Collect the current space renter state from the ViewModel
+  val spaceState by viewModel.spaceRenter.collectAsStateWithLifecycle()
+  // Trigger loading of space renter data when spaceId changes
+  LaunchedEffect(spaceId) { viewModel.getSpaceRenter(spaceId) }
```
MarcoLourenco55, line 74: Consider handling exceptions when making vm calls that may fail

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+  Column(
+      verticalArrangement = Arrangement.spacedBy(8.dp),
+      modifier = Modifier.fillMaxWidth().padding(horizontal = 25.dp)) {
+        Text(
+            "Availability",
+            style = MaterialTheme.typography.titleLarge,
+            fontWeight = FontWeight.SemiBold)
+
+        // Loop through each day's opening hours
+        openingHours.forEach { entry ->
```
MarcoLourenco55: This assumes that days are already sorted no ? Maybe consider sorting them here beforehand to make sure this is the case

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,287 @@
...
+
+// -------------------- CONTACT SECTION --------------------
+
+/**
+ * Composable that displays the contact information section of the space renter.
+ *
+ * @param spaceRenter The space renter whose contact information is displayed.
+ */
+@Composable
+fun ContactSection(spaceRenter: SpaceRenter) {
```
MarcoLourenco55, line 147: Consider not showing a given section if there is no information. For example, skip the phone part if there is no phone number, instead of displaying "- Phone: "

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Consider using account.uid for ownership equalities, instead of object's ones.

MarcoLourenco55 (changes requested):
Great PR and nice screen, thank you for the work !

A part from the comments relative to specific parts, here you can find other ones that may improve the code :
- Consider removing magic strings ("Contact", "Available spaces", "-Phone", "- Email", etc.)
- Consider using consistent colorscheme (MaterialTheme.coloScheme vs AppColors) but not both
- Consider making entire screen scrollable to avoid squeezing last part

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Fair point

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I think being bold is the right balance between "indicative" and "not too flashy". Had it been a different color it may insinuitiate that something is interactible or that it's a super important, whereas it's really just to highlight what the current day is.
When viewing UI I like to think that the most important parts should be recognizable as the most important from a far distance, having the color contrast for today's date is going to throw you off the first time you land on the screen.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Alright I'll change that.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Fixed.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Good point, I also fixed the weird `?: false` fallback you mentioned.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I'm going to implement this for the phone and website sections, since they seem to be the only 2 that are not mandatory from your `CreateSpaceRenterScreen`.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Fixed, I now compare just the uids.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Fair enough

AlexandroHattoum: Good point.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Also done.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: The VM can only throw an exception if `spaceId` is null, but it's passed as a parameter that's not null-able. Are there any other cases where the VM may fail that I am not aware of?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I can include the sorting to be safe 100% but I think the assumption is valid (could be wrong).

AlexandroHattoum:
> Great PR and nice screen, thank you for the work !
> 
> A part from the comments relative to specific parts, here you can find other ones that may improve the code :
> 
> * Consider removing magic strings ("Contact", "Available spaces", "-Phone", "- Email", etc.)
> * Consider using consistent colorscheme (MaterialTheme.coloScheme vs AppColors) but not both
> * Consider making entire screen scrollable to avoid squeezing last part

1. I think in general magic strings only apply if the string is long or it reflects a piece of context that is shared (both cases where the string is only used once or used many times). Say we were doing navigation with screen names for example, then you would absolutely have to put the strings inside variables.
But in the examples you're pointing out here, these are just UI labels and replacing them with variables would just make it harder to follow the code, don't you think?

2. I've searched with `Ctrl + F` my code but I haven't been able to find a place where I call `MaterialTheme.colorScheme`. Could you point out where you saw it?

3. Made the entire screen scrollable as requested, it's much cleaner.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Mhhh I thought that it was possible to have a connection problem with the backend and receive a null answer. But maybe I am wrong

MarcoLourenco55:
> 1. I think in general magic strings only apply if the string is long or it reflects a piece of context that is shared (both cases where > the string is only used once or used many times). Say we were doing navigation with screen names for example, then you would > absolutely have to put the strings inside variables.
> But in the examples you're pointing out here, these are just UI labels and replacing them with variables would just make it harder > to follow the code, don't you think?
>
> 2. I've searched with `Ctrl + F` my code but I haven't been able to find a place where I call `MaterialTheme.colorScheme`. Could > you point out where you saw it?
>
> 3. Made the entire screen scrollable as requested, it's much cleaner.

1) I understand your point, but I still think that a strings that may be changed or used somewhere else should be placed as a variable. This makes future modifications much easier. Furthermore, if someone would use the same text for a screen very similar later on, it is easy to use the same. Lastly, one can find more easily the place where it is used, if there is a place where you have all the strings used for the ui. But of course, this is not mandatory and depends on how you feel about it. If you prefer to keep them inside the composables, this is fine too.

2) Yes sorry, I was talking about MatherialTheme in general vs AppTheme. You use sometimes MaterialTheme calls (```MaterialTheme.typography.titleLarge``` for example, but also calls to our AppTheme specifically). I think it would be better to keep this consistent and only use one. This is not mandatory though since it doesn't impact negatively the code.
I have a small preference to use MaterialTheme consistently, since it would be easier to switch our app theme later on by binding it to MaterialTheme's one (as already done for our actual), but do as you prefer.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I believe the popup doesn't allow for overallping opening hours. 
Now for the case where start == end. You could either evaluate it as closed or as open 24h. Currently I believe the popup will take it as open 24hours.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: My current refactor should have also removed this wonky rework.

MarcoLourenco55:
Can you also center the space renter name in the topbar ? I think this would be cleaner
image

AlexandroHattoum:
> 1. I understand your point, but I still think that a strings that may be changed or used somewhere else should be placed as a variable. This makes future modifications much easier. Furthermore, if someone would use the same text for a screen very similar later on, it is easy to use the same. Lastly, one can find more easily the place where it is used, if there is a place where you have all the strings used for the ui. But of course, this is not mandatory and depends on how you feel about it. If you prefer to keep them inside the composables, this is fine too.
> 2. Yes sorry, I was talking about MatherialTheme in general vs AppTheme. You use sometimes MaterialTheme calls (`MaterialTheme.typography.titleLarge` for example, but also calls to our AppTheme specifically). I think it would be better to keep this consistent and only use one. This is not mandatory though since it doesn't impact negatively the code.
>    I have a small preference to use MaterialTheme consistently, since it would be easier to switch our app theme later on by binding it to MaterialTheme's one (as already done for our actual), but do as you prefer.

1. I think the main argument against it is just code readability, if the variable name is longer than the actual string, that's counterintuitive to me.
2. AppColors is just an easier shorthand to match the Figma design, I think most of us would rather use it than do the mapping in our heads😅

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Yes in fact, but if one day someone changes it, then you must have a problem. But fair enough, let's take this assumption as always true here

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -274,3 +273,45 @@ fun AvailabilitySection(openingHours: List<OpeningHours>) {
...
+          }
+
+      // Center Title, constrained so it NEVER touches icons
+      Text(
+          text = text,
+          style = MaterialTheme.typography.titleLarge,
+          maxLines = Int.MAX_VALUE, // Allows vertical expansion when text is long
+          modifier =
+              Modifier.align(Alignment.Center)
+                  .padding(horizontal = 48.dp) // <-- Critical fix
```
MarcoLourenco55: I think you can remove some comments, like this one "<-- Critical fix". They don't seem to add any meaningful information for the future reader

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: I researched about it, and it doesn't seem to answer a null in case of a bad connection. You can keep it like this then !

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -0,0 +1,317 @@
...
+                        modifier =
+                            Modifier.weight(1f)
+                                .testTag(
+                                    "${SpaceRenterTestTags.SPACE_RENTER_DAY_PREFIX}${entry.day}_HOURS"))
+                  }
+            }
+      }
+}
+
+@Composable
```
MarcoLourenco55, line 285: Consider adding docstrings

MarcoLourenco55 (changes requested):
Thank you very much for the changes, it renders the code much more readable !
A part from a small changes to perform from comments and docstrings, the rest seems ready to be merged for me :D

MarcoLourenco55 (approved):
Perfect, thank you again for the changes, it seems to be ready for merging to me. Nice work !

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/139/commits/bf34ee0183a3834d29a2364fa3fba0cd1c1550b2

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/139/commits/bf34ee0183a3834d29a2364fa3fba0cd1c1550b2

PR #140, opened 2025-11-12
Feat/end to end tests

### LINKED ISSUE #100: Create End to End tests

EnesKiper:
This PR is an end to end test showcasing the different features added during the sprints 3, 4, 5.
This end to end test tests the flow of one main user and simulates the behavior of other users via the back end repositories.

**Tested flows :**
* Start a discussion with other users
* Create a poll to ask them about the name for the store 
* Create a store with that name, list availabilities, and games to sell
* Create a post to announce that the store is live

**Backend flows simulated from the backend :**
* The singing up of other users
* Their votes in the poll
* Their reaction to the post


Atomium90 (commented):
Overall, this is a very good and complete end-to-end test. It covers well the new shop feature as well as the polling system. 
It might be interesting to also add a section for the new space renter, and perhaps one for shop modifications. Since we introduced the map feature, it could also be useful to navigate to the map and verify that the shop you created appears there.

Atomium90 (approved):
Overall, this is a very good and complete end-to-end test. It covers well the new shop feature as well as the polling system.
It might be interesting at a later point to test the map as well, but for now this is a very solid test.

PR #162, opened 2025-11-18
refactor(ui): refactor shop screen for new figma design

### LINKED ISSUE #154: Refactor ShopScreen
### Overview
Refactor ShopScreen to match the new Figma design. Inspire your design from real world applications.

### Constraints:
Information that needs to be displayed on the screen:
- Basic shop info: Shop's name, email, phone number, location and website/other link 
- Weekly time availability of the shop 
- Games that the shop posseses with their amount
- Photos of the shop (added by the owner of the shop)

MarcoLourenco55:
**Description**
This PR refactors ShopScreen to be able to show game images, and also changes other UI details for better visualization
It introduces the following features:
* Game images
* Game images grid of 8 games
* Pager to slide from grid to grid
* OpeningHours for all days dialog

**Visuals**
       src="https://github.com/user-attachments/assets/127cee4f-84ed-48d8-8e70-2028cff6f7e7">

**References**
* Issues: https://github.com/Meeple-Meet/Meeple-Meet/issues/154

**TODOS**
* Visualize it with true images when available using the API (yet to be implemented)
* Add the shop images composables when available (yet to be implemented)


app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
```diff
@@ -149,37 +191,42 @@ fun ShopDetails(shop: Shop, modifier: Modifier = Modifier) {
...
 
-        // Display phone contact row
+        // Display address contact row
         ContactRow(
-            Icons.Default.Phone,
-            "$PHONE_LINE_TEXT ${shop.phone}",
-            ShopTestTags.SHOP_PHONE_TEXT,
-            ShopTestTags.SHOP_PHONE_BUTTON)
+            Icons.Default.Place,
+            shop.address.name,
```
AlexandroHattoum: We're going to have to find a way to display a human-readable address, something that does not take a lot of space (at most 1 line) yet still conveys the right information.

MarcoLourenco55: This could be nice for sure, but in my opinion, we must do it in a specific kotlin class. Such that each screen can call this class functions, instead of repeating code across screens

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,165 @@ fun EditableGameItem(
...
+) {
+  val todayCalendarValue = Calendar.getInstance().get(Calendar.DAY_OF_WEEK)
+  val todayIndex = todayCalendarValue - 1
+
+  val todayEntry = openingHours.firstOrNull { it.day == todayIndex }
+  val todayLines: List<String> =
+      todayEntry?.let { humanize(it.hours).split("\n") }
+          ?: listOf(ShopUiDefaults.StringsMagicNumbers.CLOSED)
+
+  var showFullWeek by remember { mutableStateOf(false) }
```
AlexandroHattoum, line 1371: What I personally did was having the entire row clickable to display the availability popup, it's easier when you're actually on mobile I find. Discuss with me what you think.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,165 @@ fun EditableGameItem(
...
+private fun WeeklyAvailabilityDialog(
+    openingHours: List<OpeningHours>,
+    currentDayIndex: Int,
+    dayTagPrefix: String,
+    onDismiss: () -> Unit
+) {
+  val dayNames = remember { ShopFormUi.dayNames }
+  val scrollState = rememberScrollState()
+
+  AlertDialog(
```
AlexandroHattoum: Instead of an `AlertDialog`, I used `ModalBottomSheet`, it's a popup that originates from the bottom of your screen instead of popping right in the middle. It's usually more mobile friendly I find, especially since there's no interractions that the user would do with the popup to warrant it being at the middle (contrary to the poll system for example).
Your implementation beats mine here, since it's a bit cleaner in terms of the displayed text, I just think it would be perfect if it were to use `ModalBottomSheet`.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,165 @@ fun EditableGameItem(
...
+                                modifier = Modifier.weight(1f),
+                                horizontalAlignment = Alignment.End,
+                                verticalArrangement =
+                                    Arrangement.spacedBy(Dimensions.Spacing.extraSmall)) {
+                                  lines.forEach { line ->
+                                    Text(
+                                        text = line,
+                                        style = MaterialTheme.typography.bodyMedium,
+                                        fontWeight =
+                                            if (isToday) FontWeight.Bold else FontWeight.Normal,
```
AlexandroHattoum: This doesn't make it bold it makes the current day for the text italic.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,164 @@ fun EditableGameItem(
...
+ * @param onDismiss A callback function that is invoked when the dialog is dismissed
+ */
+@Composable
+private fun WeeklyAvailabilityDialog(
+    openingHours: List<OpeningHours>,
+    currentDayIndex: Int,
+    dayTagPrefix: String,
+    onDismiss: () -> Unit
+) {
+  val dayNames = remember { ShopFormUi.dayNames }
```
AlexandroHattoum: Do you need to have this inside a `remember` scope? I feel like it's just a static list that you create once.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,164 @@ fun EditableGameItem(
...
+        Row(
+            modifier =
+                Modifier.fillMaxWidth()
+                    .padding(top = Dimensions.Spacing.small)
+                    .testTag("${dayTagPrefix}${todayIndex}_HOURS"),
+            verticalAlignment = Alignment.Top) {
+
+              // Left label
+              Text(
+                  text = "Today:",
```
AlexandroHattoum: After last week's changes, this should be in the UI object for strings.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,164 @@ fun EditableGameItem(
...
+  var showFullWeek by remember { mutableStateOf(false) }
+
+  Column(
+      verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.small),
+      modifier = Modifier.fillMaxWidth()) {
+
+        // Header
+        Row(verticalAlignment = Alignment.CenterVertically, modifier = Modifier.fillMaxWidth()) {
+          Text(
+              text = "Availability",
```
AlexandroHattoum: After last week's changes, this should be in the UI object for strings.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,164 @@ fun EditableGameItem(
...
+                    .heightIn(max = Dimensions.ContainerSize.alertDialogHeight)
+                    .verticalScroll(scrollState)) {
+              Column(modifier = Modifier.fillMaxWidth()) {
+                openingHours
+                    .sortedBy { it.day }
+                    .forEach { entry ->
+                      val day = entry.day
+                      val isToday = day == currentDayIndex
+                      val dayName = dayNames.getOrNull(day) ?: "Day ${day + 1}"
+                      val lines = humanize(entry.hours).split("\n")
```
AlexandroHattoum: I notice you do this operation twice in this file (humanize(<list>).split("\n")). There's nothing particularily wrong with your approach but refactoring to a private function you can use twice is better, just for the sake of consistency accross the app (changing one but forgetting to change the other for example).

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
```diff
@@ -45,20 +64,37 @@ object ShopTestTags {
...
 
-private val horizontalPadding = Dimensions.ComponentWidth.spaceLabelWidth
+private val STOCK_BUBBLE_SIZE = 32.dp
+private val STOCK_BUBBLE_TOP_PADDING = (STOCK_BUBBLE_SIZE / 2)
+private const val GAME_NAME_MAX_LINES = 2
+private const val NOT_SHOWING_STOCK_MIN_VALUE = 0
+
+private const val GAME_IMG_RELATIVE_WIDTH = 0.85f
+private const val GAME_IMG_DEFAULT_ASPECT_RATIO = 0.75f
+const val MAX_STOCK_SHOWED = 99
```
AlexandroHattoum, line 92: I am targeting the whole section above with my comment. 
For your constants, it would be more practical to "group" them with an object, so everything related to the games would be in their own Game object, same idea for the pager/stock. You could have a misc (short for miscellanious) for things that are hard to group.

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
```diff
@@ -196,142 +243,216 @@ fun ContactSection(shop: Shop) {
...
+fun GameImageListSection(
+    games: List<Pair<Game, Int>>,
+    modifier: Modifier = Modifier,
+    clickableGames: Boolean = false,
+    title: String,
+    onClick: (Game) -> Unit = {},
 ) {
-  val daysOfWeek = remember { DateFormatSymbols().weekdays }
-  val currentDay = Calendar.getInstance().get(Calendar.DAY_OF_WEEK)
+  val clampedGames = remember(games) { games.shuffled().take(MAX_GAMES) }
```
AlexandroHattoum: I'm honestly not sure if it's a good idea to shuffle the games, normally users would expect some sort of order when viewing a list, even if it's not entirely loaded at first. 
I think just sort them either per alphabetical order (ease of navigation) or by the amount of copies of a game that exist (appeal to popularity/interest the customer).

app/build.gradle.kts
```diff
@@ -213,6 +213,7 @@ dependencies {
   debugImplementation(libs.androidx.ui.test.manifest)
   implementation(libs.material)
   implementation(libs.androidx.material.icons.extended)
+    implementation("io.coil-kt:coil-compose:2.6.0")
```
AlexandroHattoum: Sonar had a lot of things to say about this line you've added here, I'll let you take a more detailed look to see if it's pointing out a potential issue:
https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=162&id=Meeple-Meet_Meeple-Meet&open=AZqZDAN5dNIEiTHOVl1B

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1300,3 +1309,164 @@ fun EditableGameItem(
...
+ *
+ * @param openingHours The list of [OpeningHours] entries for the shop, using 0-based day indices
+ * @param dayTagPrefix The prefix used for test tags associated with day rows and navigation
+ */
+@Composable
+fun AvailabilitySection(
+    openingHours: List<OpeningHours>,
+    dayTagPrefix: String = ShopComponentsTestTags.SHOP_DAY_PREFIX
+) {
+  val todayCalendarValue = Calendar.getInstance().get(Calendar.DAY_OF_WEEK)
```
AlexandroHattoum: Sonar here suggests *I think* to use `Calendar.getInstance()[Calendar.DAY_OF_WEEK]`, but really do as you wish, it's the same logic and behavior.
https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=162&id=Meeple-Meet_Meeple-Meet&open=AZqZDANYdNIEiTHOVl1A

AlexandroHattoum (changes requested):
### Overall
Good UI refactor, matches the Figma pretty well and the app's structure better. I enjoyed reading your code, it was easy to follow for the most part. Solid work overall.

- Testing: Nothing for me to say, you've almost achieved 100% line coverage haha.
- Code: I've left a couple of comments for potential changes, nothing drastic but I advise you to take a look. Also consider looking at the feedback from sonarcloud on your code.
- UI: Your UI is clean but we have our differences for the components that both our screens uses, which is why I propose that we resolve them through this thread. We'll then be able to use the same components for more consistency through the app. (request photos/videos from me if my explanations of my work were unclear/incomplete).

(FYI: reminder to update the video you've attached to your PR as I can already see minor changes between the video and the app. Example: "Closed" text on availability section in the video is on the right, it's to the left on the app (which I think is better).

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Both are fine I guess. Putting an icon as in the figma, but then having the all row clickable, may be counter-intuitive. However, as you said, it may be easier to use in a cellphone.

I'll update it with your idea

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
MarcoLourenco55: The thing is that in our actual screen, it would be impossible to show all the available stock for a given shop, if they have for example >= 1000 games.

Knowing that, this screen should serve as a "shop front" in a sense, where users can see what the shop has to offer.

By showing always the same games (alphabetic order, amount of copies that changes slightly only, etc) the shop front wouldn't be very interesting in my opinion.
Shop may even ask why we don't show specific games in their screen in my opinion

The appeal part could interesting to dig, but I think the "number of copies" is a not necessarily a good proxy. One could maybe, later on, create a section where shops could select the hottest games at the moment for them, and then putting those in priority.

For the moment, I think making it random, makes sure that every game has the same odds to be selected and showned. And makes the shop front more like a "discover" part, than a static one would.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Yes, you are right. I think a bottom sheet may be better in this scenario

Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/e115b042a9eeaa4fe86a8654152dd94c776888b5

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/e115b042a9eeaa4fe86a8654152dd94c776888b5

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/3768b563e76eb07c12d2cfc5dca4aeb90bdf99d3

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/3768b563e76eb07c12d2cfc5dca4aeb90bdf99d3

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/bb8dc28bf1cc6aaee563329de712fb7e18cfa8b2

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
AlexandroHattoum: Fair point, leave it randomly shuffled for the moment.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1394,79 +1411,95 @@ fun AvailabilitySection(
...
+            Spacer(Modifier.height(Dimensions.Spacing.medium))
+
+            Row(
+                modifier =
+                    Modifier.fillMaxWidth()
+                        .padding(
+                            bottom = Dimensions.Spacing.medium, end = Dimensions.Spacing.medium),
+                horizontalArrangement = Arrangement.End) {
+                  TextButton(onClick = onDismiss) {
+                    Text(ShopUiDefaults.StringsMagicNumbers.ALERTDIALOG_CONFIRM_BUTTON_TEXT)
```
AlexandroHattoum: Change the `ALERTDIALOG` prefix since it's no longer an alert dialog.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
AlexandroHattoum: image
Can you make it not Italic? It's bold now but still weirdly italic.

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
```diff
@@ -149,37 +191,42 @@ fun ShopDetails(shop: Shop, modifier: Modifier = Modifier) {
 fun ContactSection(shop: Shop) {
```
AlexandroHattoum: Since we're going to share this composable, we have 2 options because our `ContactSection` composables take different parameters.
- The easy option is to just take 5 `String` arguments, each one for a contact row.
- Modify the backend to make both `Shop` and `SpaceRenter` inherit from a shared interface (e.g call it `ContactInfo`) the parameters that they share. You'd then pass to the composable a `ContactInfo` which is just a Shop/SpaceRenter with an adapter function to keep only the fields of interest for this composable. This is however **too much hastle** for a simple UI change, not to mention that both dataclasses are not in the same package (you might therefore have to create a new file just for this one feature) and not to mention the potential breaking of existing tests (shouldn't happen in theory, but I would not wish more testing related work on anyone).

Go with the easy solution if that's all the composable will take as parameters, nothing more. If it needs to take additional parameters, consider wrapping all the shop/space renter information into a UI-only dataclass and pass that as a single argument. I've shared the second solution only to outline why it is *not* a good idea.
An example for the **UI-only** data class:
```kt
data class ContactUiModel(
    val name: String,
    val address: String,
    val email: String,
    val phone: String?,
    val website: String?
)

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
AlexandroHattoum: I've come to the conclusion that a helper function will be too much of a hastle to make clean, human-readable addresses. We should look into the location APIs we're already using in the app and see if they have anything that could help.

AlexandroHattoum: For the moment, I'm just doing this in my `ContactRow` composable:
Text(
            text,
            modifier = Modifier.weight(1f).testTag(testTag),
            maxLines = 1, // this is the important line that just sets a max number of lines
            style = MaterialTheme.typography.bodyMedium,
            overflow = androidx.compose.ui.text.style.TextOverflow.Ellipsis) // this is new

MarcoLourenco55: The problem with the ```maxLines = 1``` is that only the first parts will be visible, which doesnt show the entire address

But I see your point: the first part is the most important one

However, not showing the address completed is strange behavior in my opinion. I think that modifying the API return or create a specialized class may be better

app/build.gradle.kts
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/062777d444ca989bc98edf7111285eb99da977d6

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/e60dc9ed6839722dc24dad7b05619a7242d91268

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/ee7f6dad378f4f4be26f797009249e8ba3b544ed

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/6bc9ac0c302f6bf7bb93c419d8217d41a335e03e

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
AlexandroHattoum: Put `maxLines = 2` if you find it's better, but I don't need to know that I am in Switzerland since I've accessed the screen from the map (just an example, I also don't need the many different translations of Switzerland). There's enough information in 1 line usually but I do agree that the google maps API would do better than our simple string cap.
We'll fix that next sprint as we've discussed in today's meeting.

app/src/androidTest/java/com/github/meeplemeet/ui/components/ShopComponentsTest.kt
AntoineBastide47, line 1: Please use the checkpoint system for your tests

AntoineBastide47 (changes requested):
Please use the checkpoint system for your tests

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: This is because in our typography type we have ```bodyMedium``` that is necessarely bold AND italic, if I understood correctly

I switched to ```titleLarge``` and it is now only bold

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/99f1b3b45cd64741c10e20977a23aec6cde62d37

app/src/androidTest/java/com/github/meeplemeet/ui/components/ShopComponentsTest.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/99f1b3b45cd64741c10e20977a23aec6cde62d37

app/src/main/java/com/github/meeplemeet/ui/shops/ShopScreen.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/dedabf9e327c71b677a03e4874c418817348dd50

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/ee68bc23df08cb85d714d35cfe1d4cf770bd2ae9

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/162/commits/e09650525323635edb79a96578c80a428d79d2c0

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
AlexandroHattoum: Here's how I made mine, since we still have a couple of small differences. If you agree with my version, I'll send you the full code. (The main differences are padding values and the click being on the full section vs just the title of the section)
studio64_7JWFieZJJH

You should be able to just copy paste what I've made into your file, albeit with a couple of naming differences (I really just called your `WeeklyAvailabilityDialog` for my popup and diverted slightly from your original code for the section without the popup.

MarcoLourenco55: I think that you can propose your changes in your PR, since it will build on mine, no ?
It may be clearer in that case, and we can discuss differences in your version in your PR

AlexandroHattoum: Sure why not, I'll put it as a shared component in `ShopComponents` once we completely agree on what to use.

AlexandroHattoum (approved):
Thank you for the UI and code changes. It all looks good to me now.

AntoineBastide47 (approved):

PR #165, opened 2025-11-19
Speed up the test suite to make using the CI less painfull and time consuming

AntoineBastide47:
Tests are now faster, more efficient and optimized thanks to better setup functions and better leveraging of the checkpoint system.

All tests in integration/ now run in 2m30 instead of 5m (-50%)
All tests in ui/ now run in 3m instead of 7m (-57%)
All tests in authentication/ run in 5s instead of 49s (-89%)


Atomium90 (approved):
That’s a solid optimization of the testing suite, all tests are passing and CI looks good. 
The scope might be a bit too large; ideally, you could have split it into separate PR (one per package) on an optimization branch, then merged that branch into main. 
But aside from that, everything looks good to me.

PR #168, opened 2025-11-19
Refactor/space renter screen

### LINKED ISSUE #155: Refactor SpaceRenterDetails Screen
### Overview
Refactor SpaceRenterDetails screen to match the new Figma design. Inspire your design from real world applications.

### Constraints:
Information that needs to be displayed on the screen:
- Basic space renter info: space renter's name, email, phone number, location and website/other link.
- Weekly time availability of the space renter. 
- Spaces that are available with the amount of seats and price for a seat.
- Photos of the available spaces for rent.

AlexandroHattoum:
### Overview
This PR refactors `SpaceRenterScreen.kt`, now being a more modernized and clean UI. 
It introduces the following changes to the composition of the screen:

- Image carousel placeholder
- Contact information is now less invasive
- Page slider for showing 3 different spaces per tab, as well as showing a min-max price range.
- Opening hours now just shows the availability for today. Weekly availability can be accessed through the chevron icon as `ModalBottomSheet`.
- Floating bottom bar for intstructions and reservations. It is **not yet** integrated with the backend, but it is tested such that later modifications should be smooth and easy.

Demo: https://github.com/user-attachments/assets/f822876f-dd13-4960-87ed-437fd65d0e70

### Known Limitations
- Screen does not implement the actual image gallery composable.

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
@@ -25,7 +25,7 @@ data class Game(
     val maxPlayers: Int,
     val recommendedPlayers: Int?,
     val averagePlayTime: Int?,
-    val minAge: Int?,
+    val minAge: Int? = 3,
     val genres: List<String> = emptyList()
 )
 
--- a/app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
@@ -25,7 +25,7 @@ data class Game(
     val maxPlayers: Int,
     val recommendedPlayers: Int?,
     val averagePlayTime: Int?,
-    val minAge: Int? = 3,
+    val minAge: Int? = 0,
     val genres: List<String> = emptyList()
 )
 
--- a/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
@@ -1458,16 +1458,12 @@ fun AvailabilitySectionWithChevron(
   val todayHours = openingHours.firstOrNull { it.day == today }
 
   val todayText =
-      when {
-        todayHours == null || todayHours.hours.isEmpty() -> SpaceRenterUi.Misc.NO_TIME
-        todayHours.hours.size == 1 -> {
-          val (start, end) = todayHours.hours.first()
+      if (todayHours == null || todayHours.hours.isEmpty()) {
+        listOf(SpaceRenterUi.Misc.NO_TIME)
+      } else {
+        todayHours.hours.map { (start, end) ->
           SpaceRenterUi.AvailabilitySection.timeRange(start, end)
         }
-        else ->
-            todayHours.hours.joinToString(" ") { (start, end) ->
-              SpaceRenterUi.AvailabilitySection.timeRange(start, end)
-            }
       }
 
   Column(
@@ -1475,28 +1471,47 @@ fun AvailabilitySectionWithChevron(
           Modifier.fillMaxWidth()
               .padding(horizontal = if (addPadding) Dimensions.Padding.xxLarge else 0.dp)) {
         Row(
+            horizontalArrangement = Arrangement.SpaceBetween,
+            verticalAlignment = Alignment.CenterVertically,
             modifier =
                 Modifier.fillMaxWidth()
                     .clickable { showSheet = true }
                     .testTag(SpaceRenterTestTags.AVAILABILITY_HEADER)
-                    .padding(vertical = Dimensions.Padding.medium),
-            horizontalArrangement = Arrangement.SpaceBetween,
-            verticalAlignment = Alignment.CenterVertically) {
-              Column {
+                    .padding(vertical = Dimensions.Padding.medium)) {
+              Column(modifier = Modifier.weight(1f)) {
                 Text(
                     text = SpaceRenterUi.AvailabilitySection.TITLE,
                     style = MaterialTheme.typography.titleLarge,
                     fontWeight = FontWeight.SemiBold)
-                Text(
-                    text = SpaceRenterUi.AvailabilitySection.todayDate(todayText),
-                    modifier = Modifier.testTag(ShopUiDefaults.StringsMagicNumbers.TODAY_TEXT),
-                    style = MaterialTheme.typography.bodyMedium)
+                Row(
+                    modifier =
+                        Modifier.fillMaxWidth()
+                            .padding(top = Dimensions.Padding.small)
+                            .testTag("${dayTagPrefix}${today}_HOURS"),
+                    verticalAlignment = Alignment.Top) {
+                      // Left label
+                      Text(
+                          text = SpaceRenterUi.AvailabilitySection.TODAY,
+                          style = MaterialTheme.typography.bodyMedium)
+
+                      Spacer(modifier = Modifier.width(Dimensions.Padding.medium))
+
+                      // Right – multiline times stacked neatly
+                      Column(
+                          verticalArrangement =
+                              Arrangement.spacedBy(Dimensions.Spacing.extraSmall)) {
+                            todayText.forEach { line ->
+                              Text(text = line, style = MaterialTheme.typography.bodyMedium)
+                            }
+                          }
+                    }
               }
 
               Icon(
                   Icons.Default.ChevronRight, contentDescription = null, tint = AppColors.textIcons)
             }
       }
+
   if (showSheet) {
     WeeklyAvailabilityDialog(
         openingHours = openingHours,
@@ -1534,6 +1549,8 @@ fun ContactSection(
         Text(
             text = name,
             style = MaterialTheme.typography.titleLarge,
+            maxLines = 1,
+            overflow = TextOverflow.Ellipsis,
             fontWeight = FontWeight.SemiBold,
         )
 
--- a/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
@@ -24,6 +24,7 @@ import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.NavigateNext
 import androidx.compose.material.icons.filled.Add
 import androidx.compose.material.icons.filled.Check
+import androidx.compose.material.icons.filled.ChevronRight
 import androidx.compose.material.icons.filled.Close
 import androidx.compose.material.icons.filled.Delete
 import androidx.compose.material.icons.filled.Email
@@ -62,6 +63,8 @@ import com.github.meeplemeet.model.shops.OpeningHours
 import com.github.meeplemeet.model.shops.Shop
 import com.github.meeplemeet.model.shops.ShopSearchViewModel
 import com.github.meeplemeet.model.shops.TimeSlot
+import com.github.meeplemeet.ui.space_renter.SpaceRenterTestTags
+import com.github.meeplemeet.ui.space_renter.SpaceRenterUi
 import com.github.meeplemeet.ui.theme.AppColors
 import com.github.meeplemeet.ui.theme.Dimensions
 import java.time.LocalTime
@@ -140,6 +143,7 @@ object ShopComponentsTestTags {
   // Search field internals
   const val GAME_SEARCH_FIELD = "shop_game_search_field"
   const val GAME_SEARCH_CLEAR = "shop_game_search_clear"
+  const val GAME_SEARCH_PROGRESS = "shop_game_search_progress"
   const val GAME_SEARCH_MENU = "shop_game_search_menu"
   const val GAME_SEARCH_ITEM = "shop_game_search_item"
 
@@ -1349,88 +1353,6 @@ fun EditableGameItem(
 // -------------------- AVAILABILITY SECTION --------------------
 
 /**
- * A composable function that displays the availability section for a shop, showing today's opening
- * hours and providing access to the full weekly schedule
- *
- * @param openingHours The list of [OpeningHours] entries for the shop, using 0-based day indices
- * @param dayTagPrefix The prefix used for test tags associated with day rows and navigation
- */
-@Composable
-fun AvailabilitySection(
-    openingHours: List<OpeningHours>,
-    dayTagPrefix: String = ShopComponentsTestTags.SHOP_DAY_PREFIX
-) {
-  val todayCalendarValue = Calendar.getInstance()[Calendar.DAY_OF_WEEK]
-  val todayIndex = todayCalendarValue - 1
-
-  val todayEntry = openingHours.firstOrNull { it.day == todayIndex }
-  val todayLines: List<String> =
-      todayEntry?.let { splittedHumanize(it.hours) }
-          ?: listOf(ShopUiDefaults.StringsMagicNumbers.CLOSED)
-
-  var showFullWeek by remember { mutableStateOf(false) }
-
-  Column(
-      verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.small),
-      modifier = Modifier.fillMaxWidth()) {
-
-        // Header
-        Row(
-            verticalAlignment = Alignment.CenterVertically,
-            modifier =
-                Modifier.fillMaxWidth()
-                    .clickable { showFullWeek = true }
-                    .testTag("${dayTagPrefix}NAVIGATE")) {
-              Text(
-                  text = ShopUiDefaults.StringsMagicNumbers.AVAILABILITY_SECTION_TEXT,
-                  style = MaterialTheme.typography.titleLarge,
-                  fontWeight = FontWeight.SemiBold)
-
-              Spacer(modifier = Modifier.weight(1f))
-
-              Icon(
-                  imageVector = Icons.AutoMirrored.Filled.NavigateNext,
-                  contentDescription = "Show full week opening hours")
-            }
-
-        // Today line(s)
-        Row(
-            modifier =
-                Modifier.fillMaxWidth()
-                    .padding(top = Dimensions.Spacing.small)
-                    .testTag("${dayTagPrefix}${todayIndex}_HOURS"),
-            verticalAlignment = Alignment.Top) {
-
-              // Left label
-              Text(
-                  text = ShopUiDefaults.StringsMagicNumbers.TODAY_TEXT,
-                  style = MaterialTheme.typography.bodyMedium,
-              )
-
-              Spacer(Modifier.width(Dimensions.Spacing.medium))
-
-              // Opening hours, aligned with the first one
-              Column(verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.extraSmall)) {
-                todayLines.forEach { line ->
-                  Text(
-                      text = line,
-                      style = MaterialTheme.typography.bodyMedium,
-                  )
-                }
-              }
-            }
-      }
-
-  if (showFullWeek) {
-    WeeklyAvailabilityDialog(
-        openingHours = openingHours,
-        currentDayIndex = todayIndex,
-        dayTagPrefix = dayTagPrefix,
-        onDismiss = { showFullWeek = false })
-  }
-}
-
-/**
  * A composable function that displays a bottom sheet with the full weekly opening hours for a shop
  *
  * @param openingHours The list of [OpeningHours] entries to display
@@ -1440,7 +1362,7 @@ fun AvailabilitySection(
  */
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-private fun WeeklyAvailabilityDialog(
+fun WeeklyAvailabilityDialog(
     openingHours: List<OpeningHours>,
     currentDayIndex: Int,
     dayTagPrefix: String,
@@ -1521,6 +1443,67 @@ private fun WeeklyAvailabilityDialog(
                 }
           }
         }
+  }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun AvailabilitySection(
+    openingHours: List<OpeningHours>,
+    dayTagPrefix: String = ShopComponentsTestTags.SHOP_DAY_PREFIX,
+    addPadding: Boolean = false
+) {
+  var showSheet by remember { mutableStateOf(false) }
+
+  val today = Calendar.getInstance()[Calendar.DAY_OF_WEEK] - 1
+  val todayHours = openingHours.firstOrNull { it.day == today }
+
+  val todayText =
+      when {
+        todayHours == null || todayHours.hours.isEmpty() -> SpaceRenterUi.Misc.NO_TIME
+        todayHours.hours.size == 1 -> {
+          val (start, end) = todayHours.hours.first()
+          SpaceRenterUi.AvailabilitySection.timeRange(start, end)
+        }
+        else ->
+            todayHours.hours.joinToString(" ") { (start, end) ->
+              SpaceRenterUi.AvailabilitySection.timeRange(start, end)
+            }
+      }
+
+  Column(
+      modifier =
+          Modifier.fillMaxWidth()
+              .padding(horizontal = if (addPadding) Dimensions.Padding.xxLarge else 0.dp)) {
+        Row(
+            modifier =
+                Modifier.fillMaxWidth()
+                    .clickable { showSheet = true }
+                    .testTag(SpaceRenterTestTags.AVAILABILITY_HEADER)
+                    .padding(vertical = Dimensions.Padding.medium),
+            horizontalArrangement = Arrangement.SpaceBetween,
+            verticalAlignment = Alignment.CenterVertically) {
+              Column {
+                Text(
+                    text = SpaceRenterUi.AvailabilitySection.TITLE,
+                    style = MaterialTheme.typography.titleLarge,
+                    fontWeight = FontWeight.SemiBold)
+                Text(
+                    text = SpaceRenterUi.AvailabilitySection.todayDate(todayText),
+                    modifier = Modifier.testTag(ShopUiDefaults.StringsMagicNumbers.TODAY_TEXT),
+                    style = MaterialTheme.typography.bodyMedium)
+              }
+
+              Icon(
+                  Icons.Default.ChevronRight, contentDescription = null, tint = AppColors.textIcons)
+            }
+      }
+  if (showSheet) {
+    WeeklyAvailabilityDialog(
+        openingHours = openingHours,
+        currentDayIndex = today,
+        dayTagPrefix = dayTagPrefix,
+        onDismiss = { showSheet = false })
   }
 }
 
--- a/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
@@ -1447,7 +1447,7 @@ fun WeeklyAvailabilityDialog(
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun AvailabilitySection(
+fun AvailabilitySectionWithChevron(
     openingHours: List<OpeningHours>,
     dayTagPrefix: String = ShopComponentsTestTags.SHOP_DAY_PREFIX,
     addPadding: Boolean = false
@@ -1518,10 +1518,19 @@ fun AvailabilitySection(
  * @param website website link to display
  */
 @Composable
-fun ContactSection(name: String, address: String, email: String, phone: String, website: String) {
+fun ContactSection(
+    name: String,
+    address: String,
+    email: String,
+    phone: String,
+    website: String,
+    addPadding: Boolean = false
+) {
   Column(
       verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.medium),
-      modifier = Modifier.fillMaxWidth()) {
+      modifier =
+          Modifier.fillMaxWidth()
+              .padding(horizontal = if (addPadding) Dimensions.Padding.xxLarge else 0.dp)) {
         Text(
             text = name,
             style = MaterialTheme.typography.titleLarge,
--- a/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
@@ -21,7 +21,6 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.automirrored.filled.NavigateNext
 import androidx.compose.material.icons.filled.Add
 import androidx.compose.material.icons.filled.Check
 import androidx.compose.material.icons.filled.ChevronRight
--- a/app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
@@ -476,7 +476,7 @@ private fun SpaceCard(space: Space, index: Int, isSelected: Boolean, onClick: ()
                     .testTag(SpaceRenterComponentsTestTags.SPACE_ROW_PREFIX + index),
             verticalAlignment = Alignment.CenterVertically,
             horizontalArrangement = Arrangement.SpaceBetween) {
-              Column(verticalArrangement = Arrangement.spacedBy(2.dp)) {
+              Column(verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.extraSmall)) {
                 Text(
                     text = SpaceRenterUi.SpaceSection.setSpaceNumber(index),
                     style = MaterialTheme.typography.titleMedium,
@@ -508,14 +508,17 @@ private fun SpaceCard(space: Space, index: Int, isSelected: Boolean, onClick: ()
 private fun PriceRangeChip(minPrice: Double, maxPrice: Double) {
   Row(
       verticalAlignment = Alignment.CenterVertically,
-      horizontalArrangement = Arrangement.spacedBy(4.dp)) {
-        Surface(shape = CircleShape, color = Color.Transparent, modifier = Modifier.size(24.dp)) {
-          Icon(
-              Icons.Default.Info,
-              contentDescription = null,
-              tint = AppColors.textIcons,
-              modifier = Modifier.padding(4.dp))
-        }
+      horizontalArrangement = Arrangement.spacedBy(Dimensions.Spacing.small)) {
+        Surface(
+            shape = CircleShape,
+            color = Color.Transparent,
+            modifier = Modifier.size(Dimensions.Spacing.xxLarge)) {
+              Icon(
+                  Icons.Default.Info,
+                  contentDescription = null,
+                  tint = AppColors.textIcons,
+                  modifier = Modifier.padding(Dimensions.Padding.small))
+            }
 
         Text(
             if (minPrice == maxPrice) "${minPrice}\$" else "${minPrice}-${maxPrice}\$",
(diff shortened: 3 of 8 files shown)


app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+                    }
+                    // Code needed to fix scrolling issues, otherwise UI looks weird when going from
+                    // tab to tab with less than 3 items
+                    repeat(missing) {
+                      Spacer(Modifier.height(SpaceRenterUi.SpaceSection.CARD_HEIGHT).fillMaxWidth())
+                    }
+                  }
+            }
+
+        // Pager dots
```
MarcoLourenco55, line 349: image
In my opinion it would be better to put the selected page bullet as orange (AppColors.focus). It makes it much clearer in which page we are, and it is aligned with our nomenclature (focus for parts that are focused ahahah)

What do you think ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+/**
+ * Composable that displays a card representing a space with its details.
+ *
+ * @param space The space with data (seats and price) to display.
+ * @param index The index of the space in the list.
+ * @param isSelected Whether the card is currently selected.
+ * @param onClick Callback invoked when the card is clicked.
+ */
+@Composable
+private fun SpaceCard(space: Space, index: Int, isSelected: Boolean, onClick: () -> Unit) {
```
MarcoLourenco55, line 404: (nitpick) Have you tried and considered to not use a grey card for the space cards, but instead using a small elevation ? I know that your actual implementation follows the figma design, but it could be interesting to see how it renders visually. I suspect that it could be quite appealing visually.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+              val missing = spacesPerPage - (end - start)
+              Column(
+                  modifier = Modifier.fillMaxWidth().fillMaxHeight(),
+                  verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.large)) {
+                    for (i in start until end) {
+                      SpaceCard(
+                          space = spaces[i],
+                          index = i,
+                          isSelected = selectedIndex == i,
+                          onClick = { onSelect(if (selectedIndex == i) null else i) })
```
MarcoLourenco55: When clicking on a specific space card, it changes the bottom button, which is the expected result. However, the new one slightly overlaps the pager bubbles, as you can see here :
image

Consider adjusting the space, or the button size to avoid this

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+        modifier =
+            Modifier.fillMaxWidth()
+                .clickable { showSheet = true }
+                .testTag(SpaceRenterTestTags.AVAILABILITY_HEADER)
+                .padding(vertical = Dimensions.Padding.medium),
+        horizontalArrangement = Arrangement.SpaceBetween,
+        verticalAlignment = Alignment.CenterVertically) {
+          Column {
+            Text(
+                text = SpaceRenterUi.AvailabilitySection.TITLE,
```
MarcoLourenco55: Consider putting the same format for the "Availability" header in both the screen and the bottom sheet. 
Actually, one is bold, outlined and with ":" and the other is only bold
image


I would even prefer not to have another time the header in the pop up, but do as you think is cleaner visually

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+
+/**
+ * Composable that displays a row showing today's availability and opens a bottom popup with full
+ * availability when clicked.
+ *
+ * @param openingHours List of opening hours for each day of the week.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun AvailabilityRowPopup(openingHours: List<OpeningHours>) {
```
MarcoLourenco55: I remarked that you use the hh:mm format for opening hours, instead of the am/pm.
Consider using the same formatting than other screens.

However, if you think your version is better, don't hesitate to discuss it further. We can also discuss with other persons of the group.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+ * @param maxPrice The maximum price of the spaces.
+ */
+@Composable
+private fun PriceRangeChip(minPrice: Double, maxPrice: Double) {
+  Row(
+      verticalAlignment = Alignment.CenterVertically,
+      horizontalArrangement = Arrangement.spacedBy(4.dp)) {
+        Surface(shape = CircleShape, color = Color.Transparent, modifier = Modifier.size(24.dp)) {
+          Icon(
+              Icons.Default.Info,
```
MarcoLourenco55: (nitpick) I really appreciate this small icon. However, I find it a bit counter-intuitive, since from my experience on the internet, when I see this icon I tend to click on it to have further informations. Here this is not clickable, which perturbs my brain ahahah

Could you maybe consider adding a small info bubble (as done when something is copied to the clipboard) showing additional information maybe ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+              contentDescription = null,
+              tint = AppColors.textIcons,
+              modifier = Modifier.padding(4.dp))
+        }
+
+        Text(
+            if (minPrice == maxPrice) "${minPrice}\$" else "${minPrice}-${maxPrice}\$",
```
MarcoLourenco55: Consider spacing a little bit the prices, such that it is easier to read them

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -215,38 +267,40 @@ fun ContactSection(spaceRenter: SpaceRenter) {
...
-
-        IconButton(
-            onClick = copyToClipboard,
-            modifier = Modifier.size(Dimensions.IconSize.large).testTag(buttonTag)) {
-              Icon(imageVector = icon, contentDescription = null, tint = AppColors.neutral)
-            }
+            modifier = Modifier.weight(Dimensions.Weight.full).testTag(testTag),
+            maxLines = 1,
+            style = MaterialTheme.typography.bodyMedium,
+            overflow = androidx.compose.ui.text.style.TextOverflow.Ellipsis)
```
MarcoLourenco55: Consider making imports at the start of the file

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+
+  val spacesPerPage = 3
+  val pageCount = ceil(spaces.size / spacesPerPage.toFloat()).toInt()
+  val pagerState = rememberPagerState(pageCount = { pageCount })
+
+  val minPrice = spaces.minOf { it.costPerHour }
+  val maxPrice = spaces.maxOf { it.costPerHour }
+
+  Column(
+      modifier = modifier.fillMaxWidth(), // Matches Figma spacing
```
MarcoLourenco55: Consider removing unnecessary comments

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+                    for (i in start until end) {
+                      SpaceCard(
+                          space = spaces[i],
+                          index = i,
+                          isSelected = selectedIndex == i,
+                          onClick = { onSelect(if (selectedIndex == i) null else i) })
+                    }
+                    // Code needed to fix scrolling issues, otherwise UI looks weird when going from
+                    // tab to tab with less than 3 items
+                    repeat(missing) {
```
MarcoLourenco55, line 343: One could also use the functionnality of a ```BoxWithConstraints``` to constrain the form of the box where the space cards are. A bit like it is done in ```ShopScreen.kt```. This may be a cleaner way to solve the scrolling issues you are refering to.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+            verticalAlignment = Alignment.Top,
+            state = pagerState,
+            modifier = Modifier.fillMaxWidth()) { page ->
+              val start = page * spacesPerPage
+              val end = minOf(start + spacesPerPage, spaces.size)
+              val missing = spacesPerPage - (end - start)
+              Column(
+                  modifier = Modifier.fillMaxWidth().fillMaxHeight(),
+                  verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.large)) {
+                    for (i in start until end) {
```
MarcoLourenco55: (nitpick) Consider using a functional for-loop

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+/**
+ * Composable that displays a section of available spaces in a horizontal pager format.
+ *
+ * @param spaces Carousel displaying a list of spaces to display.
+ * @param selectedIndex The index of the currently selected space, null if none is selected.
+ * @param onSelect Callback invoked upon selection/unselection of a space (click actions).
+ * @param modifier Modifier(s) to be applied to the layout.
+ */
+@Composable
+fun SpacesSection(
```
MarcoLourenco55, line 283: Consider changing the header title of the spaces sections, to not have repetitions with other sections :
image

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +355,345 @@ fun TopBarAndDivider(
...
+    spaces: List<Space>,
+    selectedIndex: Int?,
+    onSelect: (Int?) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+  if (spaces.isEmpty()) return
+
+  val spacesPerPage = 3
```
MarcoLourenco55: Consider putting this as a ```private const val``` at the beggining of the file

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I resolved this in my last commit, even though it was mostly testing centered.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: This is how it looks with 4dp of shadow elevation. Personally I neither love it nor hate it, but it feels a bit weird for a carousel item.
image

AlexandroHattoum: I've set it to 2.dp, it's fine I think now.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: In the end I've merged both of our composables for the availability section and popups, there is no extra header and the text is bigger now. I'll let you judge in my latest commit.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: This is the format now. I think AM/PM format is better for the app as our time pickers in other screens use this format.
image

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I added the icon such that the text does not look "out of place" or "thrown to the right of the UI". If you have a better idea for an icon I'm all for changing it, but currently I wouldn't know what to add in this "info bubble" you mention.
I think it's pretty intuitive that it's the price range of the spaces, if you think it's my developer bias speaking, please inform me.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Good?

image

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I now use the shared composable which handles that properly.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: The point was to not use the grey background and instead use an elevation ahahah
to see if the "card" part could be replace using the elevation instead of the grey area

But again, this is a nitpick, you can concentrate yourself in most important things

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Yes, I think this is clearer. Ty for the changes !

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Here was the issue I was having:
You have 2 pages, one with 3 space cards and one with 1-2 space cards
When you scroll between each page, the entire scrolling of the screen would reset to the latest card's "vertical offset", and that would look weird when going back to the page with 3 space cards. I had tried to use BoxWithConstraints but didn't fix it, so I resorted to this `repeat(missing)` clause. 
If you're sure that `BoxWithConstraints` would solve my issue, I'll surely try again and take inspiration from the work in `ShopScreen`.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Done

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1524,6 +1445,67 @@ private fun WeeklyAvailabilityDialog(
   }
 }
 
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun AvailabilitySectionWithChevron(
```
MarcoLourenco55, line 1450: If there are several different opening hours for today, the availability section has a strange behavior, where we see them following each other (instead of each one per line) and we don't see the icon in the right :
image

Furthermore, they are not using am/pm format

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: It's supposed to be "Available spaces", thank you for catching it, I'll fix asap.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Done.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Maybe I understood it incorrectly though, but let's see if I got it : 
you are saying that when one pager had less elements than the others, it took less space. Thus when scrolling from it to another, there is a strange scrolling behavior ?

If this is the case, don't hesitate to look at the ShopScreen, where the ```GameImageListSection``` had the same issue and was resolved using ```BoxWithConstrains```
The issue was that, one page could have 8 games, and the second only 1. Making the scrolling between the two weird. Now it is fixed with ```BoxWithConstrains``` since the second page has white spaces to compensate naturally :
image
image

If I missunderstood your problem, don't hesitate to let me know. Otherwise, don't hesitate to point any potential flows using ```BoxWithConstrains```

app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
```diff
@@ -25,7 +25,7 @@ data class Game(
     val maxPlayers: Int,
     val recommendedPlayers: Int?,
     val averagePlayTime: Int?,
-    val minAge: Int?,
+    val minAge: Int? = 3,
```
MarcoLourenco55: Why did you put this number here ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I had tried that in the Figma and thought it looked very weird. I guess I'll keep it that way for now.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Should be fixed by now, I've already commited the changes (but for forgot ro reply here)

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: I think we can let this as is then

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -271,7 +228,9 @@ fun TopBarAndDivider(
           modifier = Modifier.align(Alignment.CenterStart).width(SpaceRenterUi.ROW_WIDTH),
           verticalAlignment = Alignment.CenterVertically) {
             IconButton(onClick = onReturn) {
-              Icon(Icons.AutoMirrored.Filled.ArrowBack, contentDescription = "Back")
+              Icon(
+                  Icons.AutoMirrored.Filled.ArrowBack,
+                  contentDescription = SpaceRenterUi.IconDescriptions.BACK_BUTTON)
             }
           }
```
MarcoLourenco55, line 248: Why is the Int.MAX_VALUE used below ? Wouldn't it be a bad idea to have a HUGE number of maximal lines ?

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Thank you for trying !

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Yes thank you for the fix. However, there is a new starnge behavior that, when pushing the screen to the lowest part, and then selecting a space: one can still push it further below. The inverse behavior happens when switching from selected to unselected.
       src="https://github.com/user-attachments/assets/393033d8-157b-44bd-9123-2e880276c6bd">

(nitpick) Also the bottom buttons seem to low for me. I would maybe put a small padding from the bottom of the screen :
image

These two points may be related

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
```diff
@@ -1536,10 +1518,19 @@ private fun WeeklyAvailabilityDialog(
...
  */
 @Composable
-fun ContactSection(name: String, address: String, email: String, phone: String, website: String) {
+fun ContactSection(
+    name: String,
+    address: String,
+    email: String,
+    phone: String,
+    website: String,
+    addPadding: Boolean = false
```
MarcoLourenco55, line 1542: I guess you added the ```addPadding``` parameter to have a difference between ShopScreen and SpaceRenterScreen contact sections ?
If this is the case, consider removing this parameter and decide to one padding commun for both screens, since they use the same contact section part.

I have to admit that I prefer the version where some padding is added. However, I think ```Dimensions.Padding.xxLarge``` may be slightly too much. Consider trying smaller padding

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Perfect, thank you

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Thank you for changing it

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Thank you for the change

MarcoLourenco55 (changes requested):
Great changes that were necessary to render the SpaceRenterScreen more user friendly !

A part from comments done in specific parts, I think the rest LGTM.

(nitpick) Consider updating the PR video to show the latest screen version

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +258,302 @@ fun TopBarAndDivider(
...
+/**
+ * Composable that displays a section of available spaces in a horizontal pager format.
+ *
+ * @param spaces Carousel displaying a list of spaces to display.
+ * @param selectedIndex The index of the currently selected space, null if none is selected.
+ * @param onSelect Callback invoked upon selection/unselection of a space (click actions).
+ * @param modifier Modifier(s) to be applied to the layout.
+ */
+@Composable
+fun SpacesSection(
```
MarcoLourenco55, line 283: Sonar seems to find this section a bit too complex cognitively : https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=168&id=Meeple-Meet_Meeple-Meet&open=AZqg5b9GA33IBPfU5i7Z

Consider refactoring it a bit

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
AlexandroHattoum: I had to add it because your screen is a little different than mine, you have a lazy column and the main sections are items within that lazy column, they are all padded inwards due to that lazycolumn. 
My screen is just composables in a scaffold, each has it's own padding value (even though they're all the same). So I can't be adding padding on top of yours otherwise it would throw off your screen's structure. This section looks identical in both screens at the moment, just that 1 line of difference to make it compatible.

For the padding value I think it's fine, it's what I found is closest to the Figma design.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: When I made this I didn't think about malicious users who could have a huge name for their space_renter, even though the top bar just shows "details" now (static string).
I will however be adding a limit to the contact section, because you're right that it could be abused.

app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
AlexandroHattoum: There are tests that used to create a game through the Game() class, but due to the additional parameter they needed refactoring. I put minimum age as 3 because it's a fair choice for all board games.
If it's crumbersome or not wanted I can always fix the failing tests one by one...

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
AlexandroHattoum: Good catch, should be fixed in my latest commit.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Hopefully it passes now.

MarcoLourenco55: Still complaining unfortunately

app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
MarcoLourenco55: Which tests are you refering to ?
And, I think this is good then, but could you maybe put ```minAge: Int? = 0``` ?
This way the tests pass, and we don't constrains ourselves to board games that may have smaller ages (yes, there exists some ahahaha)

AlexandroHattoum: If there exists some then I'll just go with 0 to be safe.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
MarcoLourenco55: Unfortunately, I think you introduced a new error with your solution. In fact, you force the time range to be displayed as ```"${start}AM - ${end}PM"```
Which may not be the case, if your end is also in AM. You can have a look into the following printscreens to see a good example :
image

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -299,3 +258,302 @@ fun TopBarAndDivider(
...
+        // Pager
+        HorizontalPager(
+            verticalAlignment = Alignment.Top,
+            state = pagerState,
+            modifier = Modifier.fillMaxWidth()) { page ->
+              val start = page * SpaceRenterUi.SpaceSection.SPACES_PER_PAGE
+              val end = minOf(start + SpaceRenterUi.SpaceSection.SPACES_PER_PAGE, spaces.size)
+              val missing = SpaceRenterUi.SpaceSection.SPACES_PER_PAGE - (end - start)
+              Column(
+                  modifier = Modifier.fillMaxWidth().fillMaxHeight(),
```
MarcoLourenco55: (nitpick) Consider adding a small padding to the left and right of the space cards, such that when we swipe from page to page, we don't see them merged as in the printscreen below :
image

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
AlexandroHattoum: Well that's stupid of me, I'll fix it asap.

MarcoLourenco55 (changes requested):

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
```diff
@@ -12,55 +19,81 @@ import androidx.compose.material3.*
...
+
+    fun setSpacePrice(costPerHour: Double) = "$costPerHour\$ per hour"
+  }
+
+  object AvailabilitySection {
+    const val TITLE = "Availability"
+
+    const val TODAY = "Today:"
+
+    fun timeRange(start: String?, end: String?) = "${start}AM - ${end}PM"
```
MarcoLourenco55:
```suggestion
    private fun formatTime(raw: String?): String {
      val value = raw ?: return "-"
      val parsed = value.tryParseTime()
      return parsed?.format(TimeUi.fmt12()) ?: value
    }

    fun timeRange(start: String?, end: String?): String {
      if (start == null && end == null) return SpaceRenterUi.Misc.NO_TIME
      return "${formatTime(start)} - ${formatTime(end)}"
    }
```

MarcoLourenco55 (commented):
You can find below a suggestion's code to fix this issue : https://github.com/Meeple-Meet/Meeple-Meet/pull/168#discussion_r2546311897
 
Suggestion : https://github.com/Meeple-Meet/Meeple-Meet/pull/168#discussion_r2546455814

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: The bottom bar now sits a bit above the end of the screen, and there's no additional scrolling issues since both versions of that composable now have the same height.

app/src/main/java/com/github/meeplemeet/ui/components/ShopComponents.kt
AlexandroHattoum: Fixed thanks to your suggest.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Hopefully it passes now x2.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Yes, thank you it solved the problem.
(nitpick) However, I think the "click on a space to setup a reservation" button is too big. The bottons are also at the limit of the bottom bar, such that the bottom bar and the bottom both cut the space cards :

image

image

Can you maybe slightly adjust them to a better visualization ?

app/src/main/java/com/github/meeplemeet/model/shared/game/Game.kt
MarcoLourenco55: Thank you for adjusting it

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: I thought about your solution and think this is fine too in reality. It mimics the same behavior than the BoxWithConstrains without being to verbose

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Yes ! WP

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: Fixed in my latest commit.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: No misunderstandings, I'll check it out.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
AlexandroHattoum: I can't change the size of the "Click on a space to setup a reservation" button as that would revert the scrolling related issue. The problem before was that the screen resizes as you're already scrolled below, which leads to this small visual inconsistency. 
I originally removed the padding from the top of the bottom bar to have it take less space, I've put it's color as transparent but it seems it still cuts through some content. I'll see what I can do.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Thank you very much, it think this is much cleaner

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: Yes, I think a transparency colors for the bottom bar may be a good solution

AlexandroHattoum: Here's what I've tried so far
- The bar's background is already transparent
- I've tried wrapping the bar in a surface to mitigate scaffold effects for the elevation
- I've tried having the bottom bar outside of the scaffold's `bottomBar` argument. That now means the bar will not scroll with the user.
- Tried to implement my own bottom bar that acts like a bottom bar but to no avail, my implementation was buggy and not worth it when scaffold already provides exactly what we're looking for, just with this little caveat.
- I've tried wrapping the container around the actual bottom bar, that failed miserably.
- Tried a lot of combinations of the mentioned changes...nothing works.

The problem comes from Scaffold, it's either you choose to go along with scaffold and have this minor UI issue or implement your own bottom bar completely from scratch...and I'm not doing that xD.
If you have an idea for an easy fix I can try go ahead, this is not the only screen with this issue.

AlexandroHattoum:
> Great changes that were necessary to render the SpaceRenterScreen more user friendly !
> 
> A part from comments done in specific parts, I think the rest LGTM.
> 
> (nitpick) Consider updating the PR video to show the latest screen version

I've updated the video, it's a link now because it was too big.

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: I think you did more than I was expecting you to try ahahaha, sorry for the inconvinience!
Let's keep like this, since I would much prefer this version than the last one

Thank you very much for the changes

app/src/main/java/com/github/meeplemeet/ui/space_renter/SpaceRenterScreen.kt
MarcoLourenco55: I think putting maxLines = 1 (or 2) would be greater for the future, without changing the actual screen.
But anyways these are nitpicks, since for the moment this is not a problem

MarcoLourenco55 (approved):
Great PR, one of the cleanest screens of the app in my opinion. Sorry for all the requests, but great work solving them !

After the CI passes, you can merge 🥳

PR #173, opened 2025-11-20
fix(ui): correct usage of max bubble count

MarcoLourenco55:
This PR corrects a small bug that was found in ShopScreen. In fact, the maximum defined stock number to be displayed was not being correctly called.

Ref: issue https://github.com/Meeple-Meet/Meeple-Meet/issues/154


AlexandroHattoum (approved):
Good job finding a subtle bug like this. You can merge.

PR #189, opened 2025-11-21
 Improve keyboard-aware UX across form screens

AntoineBastide47:
## Summary

  - add KeyboardUtils, UiBehaviorConfig, and FocusableInputField so Compose forms react to IME visibility, clear focus when the keyboard hides, and expose focus state to parent scaffolds; hook these helpers into
    MainActivity and set the window to adjustResize.
  - update every create/edit/detail screen that shows a bottom action bar (auth, discussions, posts, sessions, shops, space renters) to use the new focusable fields, hide action bars when inputs are focused, add scroll
    + tap-to-dismiss behavior, and pipe focus state through LocalFocusableFieldObserver.
  - teach instrumentation helpers and E2E tests to close the keyboard explicitly and disable the new UI flags so buttons remain accessible during assertions.


Atomium90 (approved):
Thats a good addon. I let you check and fix sonar issues and that's good to me.

PR #191, opened 2025-11-22
190 bug location lost when updating item

### LINKED ISSUE #190: Bug: Location lost when updating item
**Overview**  
Updating a location item (shop, space renter, session) removes the location from the associated geopins, which causes the pin to disappear from the map.

**Steps to reproduce**  
1. Create a shop and add a location during creation  
2. Check the shop on the map → pin is visible  
3. Edit the shop without changing the location  
4. Check the map again → the shop pin has disappeared  

**Expected behavior**  
When updating an item (shop in this case), the location should remain attached to the geopin and the pin should stay visible on the map.

**Actual behavior**  
The location is removed from the associated geopin, and the pin disappears from the map.

**Additional context**  
Checking Firebase shows that the associated geopin no longer have a location after the update.

Atomium90:
**Fixes location pins disappearing from the map when updating items (shops, spaces, sessions).**

Related to #190

The issue was caused by two problems:
1. **Race condition**: Firestore `set()` could overwrite GeoFirestore fields (`g`, `l`) if geolocation write completed first
2. **Transient failures**: GeoFirestore write failures were not being retried, causing silent data loss

**Solution:**
- Changed `set()` to `set(data, SetOptions.merge())` to prevent field overwrites
- Added retry logic (3 attempts with 200ms delay) for all GeoFirestore operations

To properly test the retry behavior, I've added a `GeoFirestoreOperations` interface that wraps GeoFirestore calls. The repository uses a default implementation in production (zero behavior change), but tests can inject fake implementations to simulate failures and verify retry logic.

Tests added:
- Retry succeeds after 1-2 transient failures
- Fails correctly after 3 exhausted attempts  
- Covers both upsertGeoPin and deleteGeoPin operations

No impact on public API - `StorableGeoPinRepository()` works exactly as before.

_Generated by ClaudeAI_

### DIFF
--- a/app/src/androidTest/java/com/github/meeplemeet/integration/FirestoreGeoPinTest.kt
+++ b/app/src/androidTest/java/com/github/meeplemeet/integration/FirestoreGeoPinTest.kt
@@ -2,13 +2,18 @@
 // Then, the test suite was manually reviewed, cleaned up, and debugged.
 package com.github.meeplemeet.integration
 
+import com.github.meeplemeet.model.map.GeoFirestoreOperations
 import com.github.meeplemeet.model.map.PinType
 import com.github.meeplemeet.model.map.StorableGeoPinNoUid
+import com.github.meeplemeet.model.map.StorableGeoPinRepository
 import com.github.meeplemeet.model.shared.location.Location
 import com.github.meeplemeet.utils.FirestoreTests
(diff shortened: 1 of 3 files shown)


AlexandroHattoum (approved):
Good work fixing that bug in a clean and proper way.
Everything looks good to me, you can merge.

Issue #175, opened 2025-11-20
Discussion notifications backend


MarcoLourenco55:
### Overview
Implement backend support for notifications related to joining discussions. When a user is invited or requests to join a discussion, a notification should be created, retrievable, and markable as read/handled

### Constraints:
- Define a notification model that supports at least:
  - Type (e.g. `JOIN_DISCUSSION_REQUEST`, `JOIN_DISCUSSION_INVITE`)
  - Target user, source user, discussion id
  - Status (`UNREAD`, `READ`, `HANDLED`)
- Expose endpoints for:
  - Fetching notifications for the current user
  - Marking notifications as read
  - Accepting or rejecting join requests when applicable
- Ensure the API returns data needed by the UI (message, context, timestamps, related discussion info)
- Add basic authorization (only the target user can read/handle their notifications)$

(this text was reformulated using ChatGPT-5)

Issue #174, opened 2025-11-20
Backend for friends management


MarcoLourenco55:
### Overview
Implement the backend for the Friends feature. Users should be able to send, accept, decline, and cancel friend requests, as well as see their friends list and pending requests

### Constraints:
- Define data model(s) for friendships and friend requests (e.g. `PENDING`, `ACCEPTED`, `DECLINED`, `BLOCKED`)
- Expose endpoints for:
  - Sending a friend request.
  - Accepting / declining / cancelling a friend request
  - Listing current friends
  - Listing incoming and outgoing pending requests
- Add basic validation and permissions (cannot send request to yourself, cannot duplicate an existing pending/accepted request, etc)
- Ensure responses return all data the UI needs (ids, display name, avatar, status)
- Write basic tests for the main flows

(text reformulated by ChatGPT-5)

PR #192, opened 2025-11-23
Small Refactor

AntoineBastide47:
This PR adds a new `account` package to the `model` and`ui` packages in preparation for #175 and #174 as these will add more files to the `account` package which should not be in the seperate `auth` package


Atomium90 (approved):
LGTM

PR #125, opened 2025-11-04
Feat/shop screen

NilSamaha:
# PR: Implement ShopDetails Screen with Navigation and Tests

## Summary
This PR introduces the **ShopDetailsScreen** to the Meeple Meet app, allowing users to view detailed information about a shop, including contact info, availability, and the list of games. It also integrates this screen into the app's navigation system (with placeholders for future screens) and adds UI tests for verification.

### Screen Implementation
- Added `ShopDetailsScreen` composable with:
  - Top bar showing shop name and back button.
  - Conditional edit button visible only for shop owners.
  - Sections for:
    - **Contact info**: phone, email, address, website with copy-to-clipboard functionality.
    - **Availability**: daily opening hours with current day highlighted.
    - **Game list**: displayed in a `LazyVerticalGrid` with badges for available counts. (Game images are not yet implemented, so a generic game icon is used as a placeholder for now)
- Added supporting composables: `ShopDetails`, `ContactSection`, `AvailabilitySection`, `GameListSection`, `ContactRow`, `GameItem`.
- Added `ShopTestTags` object for UI test tagging.
- Comprehensive comments and KDocs added for clarity.

### Navigation
- Integrated `ShopDetailsScreen` into `MainActivity` navigation flow.
- Added placeholder callbacks for:
  - Edit shop screen navigation (screen not yet implemented).
- Introduced `shopId` state variable and passed `ShopViewModel` to `MeepleMeetApp` for screen-specific data.
- Navigation into the screen is not yet implemented, as it depends on the MapScreen; clicking on a map point will eventually navigate to the corresponding shop, but this feature is not available yet.

### Tests
- Added UI tests for `ShopDetailsScreen`.
- Added test tags to support robust UI testing.

## Known Limitations / TODO
- Edit screen is not yet implemented; navigation currently uses a placeholder.
- Navigation from other screens is scaffolded but requires actual screens to be connected.

## Motivation
This PR sets up the foundation for a fully functional shop detail feature, ensuring:
- Users can view detailed shop info.
- The app has a navigation structure prepared for future extensions (edit shop screen).
- UI components are testable with meaningful test tags and good test coverage.

## Testing
- Verified `ShopDetailsScreen` renders correctly with dummy shop data.
- Checked back button functionality and conditional display of edit button.
- Placeholder navigation and screen state updates confirmed via `LaunchedEffect`.

PR #193, opened 2025-11-23
feat: Implement friend system with relationship management

### LINKED ISSUE #174: Backend for friends management

AntoineBastide47:
## Summary
  - Add relationship management system supporting friend requests, friendships, and blocking
  - Implement bidirectional relationship tracking with `RelationshipStatus` enum (Friend, Sent, Pending, Blocked)
  - Create `ProfileScreenViewModel` with validation logic to prevent invalid operations
  - Add comprehensive integration tests (678 lines) covering all relationship workflows

  ## Changes
  - **Account.kt**: Add `relationships` map and helper methods (`isFriendsWith()`, `hasBlocked()`, etc.)
  - **AccountRepository.kt**: Add `sendFriendRequest()`, `acceptFriendRequest()`, `blockUser()`, `resetRelationship()` with atomic Firestore updates
  - **ProfileScreenViewModel.kt**: Implement validation layer preventing invalid operations (duplicate requests, self-operations, wrong-direction accepts)
  - **RelationshipsTests.kt**: Integration tests verifying bidirectional consistency, state transitions, complex workflows, and edge cases

  🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/model/account/Account.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/account/Account.kt
@@ -1,5 +1,7 @@
 package com.github.meeplemeet.model.account
 
+// Claude Code generated the documentation
+
 import com.github.meeplemeet.model.discussions.DiscussionPreview
 import com.github.meeplemeet.model.discussions.DiscussionPreviewNoUid
 import com.github.meeplemeet.model.discussions.fromNoUid
@@ -96,8 +98,6 @@ data class Account(
  * @property description User's self-description or bio. Null if not provided.
  * @property shopOwner Indicates whether this account owns a board game shop. Defaults to false.
  * @property spaceRenter Indicates whether this account rents out gaming spaces. Defaults to false.
- * @property relationships List of all relationships this account has with other users. Defaults to
- *   empty list.
  */
 @Serializable
 data class AccountNoUid(
@@ -107,8 +107,7 @@ data class AccountNoUid(
     val photoUrl: String? = null,
     val description: String? = null,
     val shopOwner: Boolean = false,
-    val spaceRenter: Boolean = false,
-    val relationships: List<Relationship> = emptyList()
+    val spaceRenter: Boolean = false
 )
 
 /**
--- a/app/src/main/java/com/github/meeplemeet/model/account/Account.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/account/Account.kt
@@ -62,7 +62,10 @@ enum class RelationshipStatus {
   Pending,
 
   /** Blocked - the current user has blocked the other user. */
-  Blocked
+  Blocked,
+
+  /** Deleted - removes the relationship with the other user altogether */
+  Delete,
 }
 
 /**
@@ -145,6 +148,7 @@ fun fromNoUid(
       RelationshipStatus.Sent -> sent += other
       RelationshipStatus.Pending -> pending += other
       RelationshipStatus.Blocked -> blocked += other
+      RelationshipStatus.Delete -> {} // Does not exist in the db
     }
   }
 
--- a/app/src/main/java/com/github/meeplemeet/model/account/Account.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/account/Account.kt
@@ -8,17 +8,25 @@ import kotlinx.serialization.Serializable
 /**
  * Represents a user account stored in Firestore.
  *
+ * This data class models the complete account structure including user profile information, social
+ * relationships, and discussion metadata. It serves as the runtime representation of account data
+ * throughout the application.
+ *
  * @property uid Globally unique identifier of the account (Firestore document ID).
- * @property handle unique human-readable identifier for the user
- * @property name Human-readable name of the account.
- * @property email User's email used to create the account
+ * @property handle Unique human-readable identifier for the user (e.g., @username).
+ * @property name Human-readable display name of the account.
+ * @property email User's email address used to create the account.
  * @property previews Map of discussion previews keyed by discussion ID. Each preview stores
  *   lightweight metadata (e.g. last message, unread count) for discussions this account
  *   participates in.
- * @property photoUrl User's profile picture
- * @property description A user's description about himself
- * @property shopOwner Tells if this account is a shop owner
- * @property spaceRenter Tells if this account is a space renter
+ * @property photoUrl URL to the user's profile picture. Null if no profile picture is set.
+ * @property description User's self-description or bio. Null if not provided.
+ * @property shopOwner Indicates whether this account owns a board game shop.
+ * @property spaceRenter Indicates whether this account rents out gaming spaces.
+ * @property friends List of UIDs of users who are mutual friends with this account.
+ * @property sent List of UIDs of users to whom this account has sent friend requests.
+ * @property pending List of UIDs of users who have sent friend requests to this account.
+ * @property blocked List of UIDs of users who have been blocked by this account.
  */
 data class Account(
     val uid: String,
@@ -29,13 +37,67 @@ data class Account(
     var photoUrl: String? = null,
     var description: String? = null,
     var shopOwner: Boolean = false,
-    var spaceRenter: Boolean = false
+    var spaceRenter: Boolean = false,
+    val friends: List<String> = emptyList(),
+    val sent: List<String> = emptyList(),
+    val pending: List<String> = emptyList(),
+    val blocked: List<String> = emptyList(),
+)
+
+/**
+ * Represents the status of a relationship between two users.
+ *
+ * This enum is used to categorize the type of social connection or interaction between the current
+ * user and another user in the system.
+ */
+@Serializable
+enum class RelationshipStatus {
+  /** Mutual friendship - both users have accepted the friend request. */
+  Friend,
+
+  /** Friend request sent - the current user has sent a request to the other user. */
+  Sent,
+
+  /** Friend request pending - the other user has sent a request to the current user. */
+  Pending,
+
+  /** Blocked - the current user has blocked the other user. */
+  Blocked
+}
+
+/**
+ * Represents a relationship between the current user and another user.
+ *
+ * This data class pairs a user's UID with the status of their relationship to the current account.
+ * It is used for serialization to/from Firestore.
+ *
+ * @property uid The unique identifier of the other user in the relationship.
+ * @property status The type of relationship (Friend, Sent, Pending, or Blocked). Defaults to Friend
+ *   if not specified.
+ */
+@Serializable
+data class Relationship(
+    val uid: String,
+    val status: RelationshipStatus = RelationshipStatus.Friend
 )
 
 /**
  * Minimal serializable form of [Account] without the UID, used for Firestore storage.
  *
- * Firestore stores the UID as the document ID, so it is omitted from the stored object.
+ * Firestore stores the UID as the document ID, so it is omitted from the stored object. This class
+ * is optimized for serialization and contains only the fields that need to be persisted in the
+ * database. The relationships are stored as a flat list rather than separate fields for friends,
+ * sent, pending, and blocked.
+ *
+ * @property handle Unique human-readable identifier for the user. Defaults to empty string.
+ * @property name Human-readable display name. Defaults to empty string.
+ * @property email User's email address. Defaults to empty string.
+ * @property photoUrl URL to the user's profile picture. Null if not set.
+ * @property description User's self-description or bio. Null if not provided.
+ * @property shopOwner Indicates whether this account owns a board game shop. Defaults to false.
+ * @property spaceRenter Indicates whether this account rents out gaming spaces. Defaults to false.
+ * @property relationships List of all relationships this account has with other users. Defaults to
+ *   empty list.
  */
 @Serializable
 data class AccountNoUid(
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
```diff
@@ -130,19 +189,133 @@ class AccountRepository : FirestoreRepository("accounts") {
...
-   * The flow emits a complete Account object including all discussion previews. The listeners are
-   * automatically cleaned up when the flow is cancelled.
+   * The listeners are automatically cleaned up when the flow is cancelled.
    *
    * @param accountId The ID of the account to listen to
-   * @return A Flow that emits Account objects when changes are detected
+   * @return A Flow that emits Account objects when changes are detected in the account, its
+   *   previews, or its relationships
    */
   fun listenAccount(accountId: String): Flow<Account> = callbackFlow {
```
MarcoLourenco55, line 329: Consider refactoring this function, due to its high cognitive complexity (https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=193&id=Meeple-Meet_Meeple-Meet&open=AZqxu-r9Fx5fnmI6B-Xq)

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
```diff
@@ -23,8 +24,75 @@ import kotlinx.coroutines.tasks.await
...
+      collection.document(uid).collection(Account::relationships.name)
+
+  /**
+   * Extracts discussion previews from a Firestore query snapshot.
+   *
+   * @param documents The list of documents from the previews subcollection
+   * @return A map of discussion IDs to their preview objects
+   */
+  private fun extractPreviews(
+      documents: List<com.google.firebase.firestore.DocumentSnapshot>
```
MarcoLourenco55: Consider defining imports at the start of the file

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
```diff
@@ -1,24 +1,74 @@
...
+  Sent,
+
+  /** Friend request pending - the other user has sent a request to the current user. */
+  Pending,
+
+  /** Blocked - the current user has blocked the other user. */
+  Blocked,
+
+  /** Deleted - removes the relationship with the other user altogether */
+  Delete,
```
MarcoLourenco55: I am not sure of understanding the use of this enum here (```Delete```). Since, one user that deletes another one from it's friends list should only have it removed, I don't think this enum is useful per-se. Even in your tests you never use it and only check for presence/absence of relationship

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
```diff
@@ -1,24 +1,74 @@
...
 import kotlinx.serialization.Serializable
 
+/**
+ * Represents the status of a relationship between two users.
+ *
+ * This enum is used to categorize the type of social connection or interaction between the current
+ * user and another user in the system.
+ */
+@Serializable
+enum class RelationshipStatus {
```
MarcoLourenco55, line 17: Consider using CAPS for enums (```FRIEND``` instead of ```Friend``` for example)

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
AntoineBastide47: This enum value is used only in the repository to signal that a relationship needs to be removed. It will never be used in the UI and tests. It is just defined as such so that I did not need to add extra parameters to the repository functions.

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
AntoineBastide47: I am just using the same style as in NavigationSystem.kt

app/src/main/java/com/github/meeplemeet/model/account/ProfileScreenViewModel.kt
```diff
@@ -0,0 +1,149 @@
...
+   * - The accounts are the same user
+   * - Either user has blocked the other (use blockUser to unblock instead)
+   *
+   * If validation passes, the relationship is reset asynchronously via the repository.
+   *
+   * @param account The first account in the relationship
+   * @param friend The second account in the relationship
+   */
+  fun resetRelationship(account: Account, friend: Account) {
+    if (sameOrBlocked(account, friend)) return
```
MarcoLourenco55, line 155: Following the Kdoc of this function it would be possible to unblock someone by resetting the relationship
I may have understood it incorrectly, but this line seems to make this impossible, due to the early returning

Your tests also assert for this impossibility (in ```fun viewModelResetRelationship_preventsInvalidOperations```)

Make sure the viewmodel follows the same principle than the repository

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
```diff
@@ -29,13 +79,25 @@ data class Account(
...
+    val relationships: Map<String, RelationshipStatus> = emptyMap()
 )
 
 /**
  * Minimal serializable form of [Account] without the UID, used for Firestore storage.
  *
- * Firestore stores the UID as the document ID, so it is omitted from the stored object.
+ * Firestore stores the UID as the document ID, so it is omitted from the stored object. This class
+ * is optimized for serialization and contains only the fields that need to be persisted in the
+ * database. The relationships are stored as a flat list rather than separate fields for friends,
```
MarcoLourenco55: Maybe I missunderstood what you meant by "flat list", but is this Kdoc up to date with the new version of the code ?

app/src/main/java/com/github/meeplemeet/model/account/ProfileScreenViewModel.kt
AntoineBastide47: Ah yes good catch

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
```diff
@@ -51,24 +113,40 @@ data class AccountNoUid(
...
-        photoUrl = accountNoUid.photoUrl,
-        description = accountNoUid.description,
-        shopOwner = accountNoUid.shopOwner,
-        spaceRenter = accountNoUid.spaceRenter,
-    )
+    previews: Map<String, DiscussionPreviewNoUid> = emptyMap(),
+    relationships: List<Relationship> = emptyList()
+): Account {
+  // Convert relationship list to map for efficient lookup
+  val mappedRelationships = mutableMapOf<String, RelationshipStatus>()
```
MarcoLourenco55: Consider using a functional approach here, since it is cleaner and safer : ```val mappedRelationships = relationships.associate { it.uid to it.status }```

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
```diff
@@ -49,51 +117,42 @@ class AccountRepository : FirestoreRepository("accounts") {
...
    * @param id The account ID to retrieve
-   * @return The Account object with populated discussion previews
+   * @return The Account object with populated discussion previews and relationships map
    * @throws AccountNotFoundException if the account does not exist
    */
   suspend fun getAccount(id: String): Account {
     val snapshot = collection.document(id).get().await()
     val account = snapshot.toObject(AccountNoUid::class.java) ?: throw AccountNotFoundException()
 
     val previewsSnap = collection.document(id).collection("previews").get().await()
```
MarcoLourenco55, line 133: I think replacing ```"previews``` with ```Account::previews.name``` may be more consistent with your other collection calls

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
MarcoLourenco55: I think the idiomatic way of calling it in Kotlin is using CAPS. I don't think duplicating non-idiomatic writing is a good idea, even for consistency across the project. I think it is better renaming those used elsewhere.

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
```diff
@@ -130,53 +188,295 @@ class AccountRepository : FirestoreRepository("accounts") {
...
+   *   performing the blocking action
+   */
+  suspend fun blockUser(accountId: String, otherId: String, otherBlockedAccount: Boolean) {
+    val batch = db.batch()
+
+    val aRef = relationships(accountId).document(otherId)
+    val bRef = relationships(otherId).document(accountId)
+
+    batch[aRef] = mapOf(FIELD_STATUS to RelationshipStatus.Blocked)
+    if (!otherBlockedAccount) batch.delete(bRef)
```
MarcoLourenco55: I think this may create a race condition on blocking status from both persons, if done at the same time

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
```diff
@@ -130,53 +188,295 @@ class AccountRepository : FirestoreRepository("accounts") {
...
+     * @return A ListenerRegistration that can be used to remove the listener
+     */
+    private fun listenToPreviews(
+        accountRef: com.google.firebase.firestore.DocumentReference,
+        flow: kotlinx.coroutines.channels.ProducerScope<Account>,
+        accountId: String,
+        accountNoUid: AccountNoUid
+    ): ListenerRegistration {
+      return accountRef.collection(Account::previews.name).addSnapshotListener { qs, e ->
+        if (e != null) {
```
MarcoLourenco55, line 401: Consider logging error information here and on the listening functions below

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
AntoineBastide47: This change has been made on the `feat/notification-backend` that will be in a PR very soon, so I won't address it here.

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
AntoineBastide47: Fixed

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
AntoineBastide47: Done

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
AntoineBastide47: I changed it to a transaction instead of a batch which will prevent the race condition

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
AntoineBastide47: This isn't a real "error" it only fails if either there is no connection or there are no previews. So this isn't an error since without internet, we can't do anything so return nothing and same for when there are no previews.

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
MarcoLourenco55: I see your point, but I think enums should declare "states" and not "operations"

I think it would be clearer to have it removed and use ```null``` instead in ```setFriendStatus``` and ```resetRelationship```. You can make use of ```?``` type definition to do so

MarcoLourenco55 (changes requested):
Thank you very much for introducing the friend's management backend, it will make the app much more usable !

A part from the comments and requests done in specific parts, here are some general comments that could improve the code:

*  Consider using more readable variables inside your functions, such that a reader can understand it directly. For example, in ```blockUser``` function you refer to the relationship status as ```a```, which is not human friendly. But other examples exists in the code.

* Consider making more concise and clearer commits. For example, commit https://github.com/Meeple-Meet/Meeple-Meet/pull/193/commits/b701426ca19d3411631aaa3b96a60355ed97768c, could be separated in several small ones (each solving one of the comments in this PR). Moreover, the descriptions should contain information such that one can understand quite easily what you have done

* (see comment above) Consider attaching these commits to their respective comment, such that the reviewer can easily see that it was solved and check the changes. You can find  examples in the following PRs : https://github.com/Meeple-Meet/Meeple-Meet/pull/125 and https://github.com/Meeple-Meet/Meeple-Meet/pull/162

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
AntoineBastide47: Fixed in b125badd92b3ec8d2240acb69655b676bc13a191

app/src/main/java/com/github/meeplemeet/model/account/Account.kt
AntoineBastide47: Fixed in b125badd92b3ec8d2240acb69655b676bc13a191

MarcoLourenco55 (changes requested):
Thank you for the functionality changes, I think the code is clearer now.
However, as already commented before, I think that renaming some of your variables to make them more human-readable is necessary. Please consider applying this change prior to merge.

A part from that, the code LGTM !

MarcoLourenco55 (changes requested):
There are still some places where renaming is necessary, as for example : ```blockUser```,  ```setFriendStatus```, ```listenToPreviews``` and ```listenToRelationships```

Example of such variables that could be better named are:
* ```aRef``` and ```bRef```
* ```tx```
* ```qs```
* ```e``` (this one could be discussed)

MarcoLourenco55 (changes requested):
Thank you for renaming some of these variables, but I think some of them could be renamed too. I guess they make sense for you, but for a new reader, they may not be clear enough in my opinion

Also consider naming more precisely your commits. I think something like "renaming variables" would be greater for example for this commit https://github.com/Meeple-Meet/Meeple-Meet/pull/193/commits/2f3b6d8f58ee75120becad4eddfc3d6c823baf54

AntoineBastide47:
`qs` is an abreviation for `QuerySnapshot` as that is the type returned by the listeners.
`tx` is an abreviation for `Transaction`.
`eN` is for `errorN`

So the naming is correct.

MarcoLourenco55:
I think renaming them to ```querySnap``` and ```transaction``` could be better than ```qs``` and ```tx```

Of course, renaming are always prune to reader comprehension. For example, I didn't pointed out about the ```db``` one (abbrevation for data base), since for me this is quite clear from the code and the context for a first time reader.

Maybe ```qs``` and ```tx``` are fine too, but I had to dig on the code to be able to understand them. I could not do this at first sight.

I think we should ask the opinion on a third member such that he can decide on which version he/she prefers. What do you think @AlexandroHattoum ?

AlexandroHattoum:
> I think renaming them to `querySnap` and `transaction` could be better than `qs` and `tx`
> 
> Of course, renaming are always prune to reader comprehension. For example, I didn't pointed out about the `db` one (abbrevation for data base), since for me this is quite clear from the code and the context for a first time reader.
> 
> Maybe `qs` and `tx` are fine too, but I had to dig on the code to be able to understand them. I could not do this at first sight.
> 
> I think we should ask the opinion on a third member such that he can decide on which version he/she prefers. What do you think @AlexandroHattoum ?

They're both clear enough as abbreviations. I think you have to anyways look at the variable declaration, and that will give you a good enough idea of what you're using/what you're dealing with. If it's unclear to others just add a short comment above/ to the right of the line.

MarcoLourenco55 (approved):
Since the last two renaming seem fine to @AlexandroHattoum, I have nothing more to say about the rest of the code that LGTM !

Thank you again for introducing the friend's management backend

PR #195, opened 2025-11-24
ci: Cancel previous workflows on the same ref

AntoineBastide47:
Cancel previous workflows if they are the same workflow on the same ref (branch/tags)
with the same event (push/pull_request) even if they are in progress.
This setting will help reduce the number of duplicated workflows.

[Sourced From](https://github.com/apache/kvrocks/pull/667)


Atomium90 (approved):
LGTM

PR #199, opened 2025-11-24
Map Clustering & Cluster Preview

### LINKED ISSUE #176: Implement cluster pins for the map
### Overview
Rework the pin display on the map to cluster pins that share the same location. Add a dedicated marker icon for clusters. When a cluster is tapped, open a bottom sheet showing all items at that location. The user can then select an item from the list to open it in the bottom sheet, following the current behavior

Atomium90:
Introduce a flexible **clustering system** for map pins and add **cluster preview functionality** to display aggregated pins. Refactor `MapViewModel` and tests for efficiency and clarity.

### Cluster Strategy
- Introduce `ClusterStrategy` interface for pluggable clustering algorithms.
- Implement `DistanceBasedClusterStrategy` that groups pins based on proximity and supports zoom-level-dependent clustering.

### ClusterManager
- Acts as a bridge between the ViewModel and the cluster strategy.
- Responsible for computing clusters efficiently and provides a clean API for querying clusters by zoom level.

### MapViewModel Refactor
- Optimized pin handling with **debounced updates** for `allGeoPins`.
- Cluster cache added to reduce recomputation on frequent updates.
- Add methods for managing **cluster selection and previews**: `selectCluster()`, `selectPinFromCluster()`, `clearSelectedCluster()`.
- Cluster cache is invalidated when filters or zoom changes.
- Remove direct use of `geoPins` in favor of `getClusters()`, ensuring clustering is always applied.

### UI & Previews
- Supports **preview fetching per cluster or per pin**.
- Selected cluster previews are exposed in `uiState.selectedClusterPreviews`.
- Single-pin selection clears cluster selection and vice versa.

### Tests
- Add tests covering cluster computation, pin addition/removal with radius updates, and cluster preview selection.
- Remove redundant tests that were verifying `geoPins` logic.

https://github.com/user-attachments/assets/7de7ff64-bbb8-4116-9826-d70575245492

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
@@ -371,7 +371,7 @@ class MapViewModel(
    */
   fun selectPin(pin: GeoPinWithLocation) {
     viewModelScope.launch {
-      _uiState.update { it.copy(isLoadingPreview = true, selectedGeoPin = pin.geoPin) }
+      _uiState.update { it.copy(isLoadingPreview = true) }
       try {
         val preview = markerPreviewRepo.getMarkerPreview(pin.geoPin)
         _uiState.update {
--- a/app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
@@ -13,6 +13,8 @@ import com.github.meeplemeet.model.map.cluster.DistanceBasedClusterStrategy
 import com.github.meeplemeet.model.sessions.SessionRepository
 import com.github.meeplemeet.model.shared.location.Location
 import com.google.firebase.firestore.GeoPoint
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
```diff
@@ -86,12 +217,11 @@ class MapViewModel(
...
     stopGeoQuery()
 
-    // Pre-fetch session IDs where user is participating
     viewModelScope.launch {
-      val userSessionIds = mutableSetOf<String>()
       try {
+        // Load user sessions for filtering
         val ids = sessionRepo.getSessionIdsForUser(currentUserId)
-        userSessionIds.addAll(ids)
+        userSessionIds = ids.toSet()
```
dtbcds, line 224: I couldn't find any subsequent usage (for filtering). If unused, remove it or add a TODO explaining future use.

app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
```diff
@@ -901,3 +1083,60 @@ private fun rememberMarkerIcon(
...
+ *
+ * The icon is a circle whose color depends on the cluster size (small/medium/large), and the size
+ * number is drawn centered in white text.
+ *
+ * @param size number of items in the cluster
+ * @param diameterDp diameter of the cluster icon in dp (default is 64dp)
+ * @return a [BitmapDescriptor] usable as a Google Maps marker icon
+ */
+@Composable
+private fun rememberClusterIcon(size: Int, diameterDp: Int = 32): BitmapDescriptor {
```
dtbcds, line 1098: Make the doc match the code, or set default to the documented value. (doc says default diameter is 64dp, but function default diameterDp = 32.)

dtbcds (commented):
After having tested your branch, it seems the clusters do work but since the query has a pretty short distance for loading pins, the clusters only show up for extremely close shops. Now that the clusters are implemented, consider raising this threshold?

dtbcds (changes requested):
Overall, this is a really good addition. However, you have some mismatching kdoc and unused variables/imports (as mentioned above), and I do believe the threshold of loading should be raised, do let me know otherwise if you have concerns.

app/src/main/java/com/github/meeplemeet/model/map/MapViewModel.kt
Atomium90: No it's used in the onKeyEntered functnion, for filtering

app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
Atomium90: good catch, I cahnged it but forget updating doc

dtbcds (approved):
Overall a really good PR, the implementation is really well done. Ready to be merged!

PR #201, opened 2025-11-25
Feat/session photos refactor

EnesKiper:
**Summary :**
This pull request introduces comprehensive support for session photo attachments in the app. It adds the ability to upload, retrieve, and manage photos associated with game sessions, including backend storage, frontend data models, and new repository/viewmodel logic. The changes affect several core files and extend the session data model to include images, enabling features like photo galleries per session and user-centric photo queries.

**Description :**
* SessionPhoto : I created a new class to have the uuid and the url of a photo, this is so that from clicking on a photo, we have a way to look for the session storing that photo, so we have to store a uuid too.
* ImageRepository : I added functions to save, load, delete photos from the firebase storage via the path to the session and the uuid
* SessionRepository : I created functions to store the photo obejct (uuid, url) into the firestore database in the session
* View models : I added functions calling the functions from both repositories to have consistency between the firestore databse and the firebase storage

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
@@ -424,8 +424,8 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
   /**
    * Saves session photos to Firebase Storage and local cache, returning SessionPhoto objects.
    *
-   * Generates unique UUIDs for each photo, uploads them in parallel, and fetches download URLs.
-   * The caller is responsible for updating session metadata with the returned SessionPhoto objects.
+   * Generates unique UUIDs for each photo, uploads them in parallel, and fetches download URLs. The
+   * caller is responsible for updating session metadata with the returned SessionPhoto objects.
    *
    * @param context Android context for accessing cache directory
    * @param discussionId The unique identifier for the discussion/session
@@ -436,25 +436,26 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
    * @throws RemoteStorageException if Firebase Storage upload fails
    */
   suspend fun saveSessionPhotos(
-    context: Context,
-    discussionId: String,
-    vararg inputPaths: String
+      context: Context,
+      discussionId: String,
+      vararg inputPaths: String
   ): List<SessionPhoto> = coroutineScope {
-      // Upload all images in parallel and get their URLs
-      inputPaths.map { inputPath ->
+    // Upload all images in parallel and get their URLs
+    inputPaths
+        .map { inputPath ->
           async {
-              val uuid = UUID.randomUUID().toString()
-              val path = sessionPhotoPath(discussionId, uuid)
-              saveImage(context, inputPath, path)
-              
-              // Get download URL from Firebase Storage
-              val url = storage.reference.child(path).downloadUrl.await().toString()
-              
-              SessionPhoto(uuid, url)
-          }
-      }.awaitAll()
-  }
+            val uuid = UUID.randomUUID().toString()
+            val path = sessionPhotoPath(discussionId, uuid)
+            saveImage(context, inputPath, path)
 
+            // Get download URL from Firebase Storage
+            val url = storage.reference.child(path).downloadUrl.await().toString()
+
+            SessionPhoto(uuid, url)
+          }
+        }
+        .awaitAll()
+  }
 
   /**
    * Loads session photos for the given UUIDs
@@ -467,20 +468,23 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
    * @throws RemoteStorageException if Firebase Storage operations fail
    */
   suspend fun loadSessionPhotos(
-    context: Context,
-    discussionId: String,
-    photoUuids: List<String>
+      context: Context,
+      discussionId: String,
+      photoUuids: List<String>
   ): List<Pair<String, ByteArray>> = coroutineScope {
-      // Load each image in parallel
-      val bytes = photoUuids.map { uuid ->
-          async {
-              val path = sessionPhotoPath(discussionId, uuid)
-              loadImage(context, path)
-          }
-      }.awaitAll()
-      
-      // Zip UUIDs with their bytes
-      photoUuids.zip(bytes)
+    // Load each image in parallel
+    val bytes =
+        photoUuids
+            .map { uuid ->
+              async {
+                val path = sessionPhotoPath(discussionId, uuid)
+                loadImage(context, path)
+              }
+            }
+            .awaitAll()
+
+    // Zip UUIDs with their bytes
+    photoUuids.zip(bytes)
   }
 
   /**
@@ -497,7 +501,11 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
    * @throws DiskStorageException if disk delete fails
    * @throws RemoteStorageException if Firebase Storage delete fails
    */
-  suspend fun deleteSessionPhoto(context: Context, discussionId: String, photoUuid: String): String {
+  suspend fun deleteSessionPhoto(
+      context: Context,
+      discussionId: String,
+      photoUuid: String
+  ): String {
     val path = sessionPhotoPath(discussionId, photoUuid)
     deleteImages(context, path)
     return photoUuid
--- a/app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
@@ -1,4 +1,5 @@
 package com.github.meeplemeet.model.images
+// AI was used in this file
 
 import android.content.Context
 import android.graphics.Bitmap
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
```diff
@@ -108,6 +111,11 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
 
   private fun spaceRenterPath(id: String) = "${RepositoryProvider.spaceRenters.collectionName}/$id"
 
+  private fun sessionPath(discussionId: String) = "discussions/${discussionId}/session"
```
AntoineBastide47, line 114: You can merge this function with the one bellow as it is only used there.

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
```diff
@@ -413,6 +421,96 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
...
+   * @return List of SessionPhoto objects containing UUID and URL for each uploaded photo
+   * @throws ImageProcessingException if image encoding fails
+   * @throws DiskStorageException if disk write fails
+   * @throws RemoteStorageException if Firebase Storage upload fails
+   */
+  suspend fun saveSessionPhotos(
+      context: Context,
+      discussionId: String,
+      vararg inputPaths: String
+  ): List<SessionPhoto> = coroutineScope {
```
AntoineBastide47, line 442: Is there a specific requirement that prevents you from using `ImageRepository.saveImages` which already does all of this.

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
```diff
@@ -413,6 +421,96 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
...
+   * @param photoUuids List of photo UUIDs to load
+   * @return List of pairs containing photo UUIDs and their corresponding image byte arrays
+   * @throws DiskStorageException if disk read fails
+   * @throws RemoteStorageException if Firebase Storage operations fail
+   */
+  suspend fun loadSessionPhotos(
+      context: Context,
+      discussionId: String,
+      photoUuids: List<String>
+  ): List<Pair<String, ByteArray>> = coroutineScope {
```
AntoineBastide47, line 474: Is there a specific requirement that prevents you from using `ImageRepository.loadImages` which already does all of this.

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
```diff
@@ -413,6 +421,96 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
...
+   * SessionRepository.removeSessionPhoto().
+   *
+   * @param context Android context for accessing cache directory
+   * @param discussionId The unique identifier for the discussion/session
+   * @param photoUuid The UUID of the photo to delete (without .webp extension)
+   * @return The UUID of the deleted photo (same as input, for coordination convenience)
+   * @throws DiskStorageException if disk delete fails
+   * @throws RemoteStorageException if Firebase Storage delete fails
+   */
+  suspend fun deleteSessionPhoto(
```
AntoineBastide47, line 504: Why do you return the uid of the deleted photo ? You shouldn't need it in the future

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
```diff
@@ -413,6 +421,96 @@ class ImageRepository(private val dispatcher: CoroutineDispatcher = Dispatchers.
...
+      context: Context,
+      discussionId: String,
+      photoUuid: String
+  ): String {
```
AntoineBastide47, line 508: Consider writing it like so to be more consistent with the existing delete functions:
` = deleteImages(context, sessionPhotoPath(discussionId, photoUuid))`

app/src/main/java/com/github/meeplemeet/model/sessions/Session.kt
```diff
@@ -15,12 +15,14 @@ import kotlinx.serialization.Serializable
...
  */
 @Serializable
 data class Session(
     val name: String = "",
     val gameId: String = "",
     val date: Timestamp = Timestamp.now(),
     val location: Location = Location(),
-    val participants: List<String> = emptyList()
+    val participants: List<String> = emptyList(),
+    val sessionPhotos: List<SessionPhoto> = emptyList()
```
AntoineBastide47, line 27: I might be wrong, but I think we agreed on a single memory photo for sessions. But we can discuss using multiple intead.

app/src/main/java/com/github/meeplemeet/model/sessions/SessionRepository.kt
```diff
@@ -104,7 +109,44 @@ class SessionRepository(
     while (true) {
       val query =
           discussions
-              .whereArrayContains("session.participants", userId)
+              .whereArrayContains(sessionParticipants, userId)
```
AntoineBastide47, line 112: Here you are querying all the existing discussions to find the one the user is a part of and a part of the session. Instead, you should query all the discussions the user is a part of using Account.previews that contains all the ids of the discussions the user is a part of.

app/src/main/java/com/github/meeplemeet/model/sessions/SessionRepository.kt
```diff
@@ -145,7 +187,60 @@ class SessionRepository(
    * @return Cold [Flow] that delivers a list of discussion document ids.
    */
   fun getSessionIdsForUserFlow(userId: String): Flow<List<String>> =
-      discussions.whereArrayContains("session.participants", userId).snapshots().map { snap ->
+      discussions.whereArrayContains(sessionParticipants, userId).snapshots().map { snap ->
```
AntoineBastide47, line 190: Same comment as for the non flow function.

app/src/main/java/com/github/meeplemeet/model/sessions/SessionRepository.kt
```diff
@@ -104,7 +109,44 @@ class SessionRepository(
...
+   * participant list contains [userId] and the session date is in the past.
+   *
+   * This method performs a paginated query to avoid reading too many documents at once. It pages
+   * through results using `startAfter(lastSnapshot)` and a `limit(batchSize)`.
+   *
+   * @param userId user UID to look for in session.participants.
+   * @param batchSize maximum documents to fetch per query page (default 50).
+   * @return list of discussion document IDs matching the criteria.
+   */
+  suspend fun getPastSessionIdsForUser(userId: String, batchSize: Long = 50): List<String> {
```
AntoineBastide47, line 140: #### This needs a rework. Once a session ends, it should be removed from it's discussion and added to another collection.

Like so:
`discussion/{discussionUid}/session` -> `past_sessions/{randomUUID}`
`account/{accountUid}/past_sessions/{sessionUid}` -> `past_sessions/{sessionUid}`

So from then on the function would query the subcollection past_sessions in account.

app/src/main/java/com/github/meeplemeet/model/sessions/SessionRepository.kt
```diff
@@ -145,7 +187,60 @@ class SessionRepository(
...
+   * Real-time stream of discussion **document ids** whose embedded session contains the given user
+   * in its participant list and the session date is in the future.
+   *
+   * The flow emits a new list on every relevant create / update / delete event in Firestore, making
+   * it suitable for observing a live session list.
+   *
+   * @param userId UID of the participant to filter for.
+   * @return Cold [Flow] that delivers a list of discussion document ids.
+   */
+  fun getUpcomingSessionIdsForUserFlow(userId: String): Flow<List<String>> =
```
AntoineBastide47, line 204: This can and should be optimized like stated previously

AntoineBastide47 (changes requested):
A solid start, but I think the description of your task was missleading so changes are required. I will convert your PR to a draft so as to not use up Blacksmith resources (I am not dismissing it, do not worry). Once it is ready for review, undraft it.

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
EnesKiper: Yes you're right, i was using it somewhere else and removed it

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
EnesKiper: You're right, i don't even use it, i forgot to remove it

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
EnesKiper: I had to pass a specific uuid since i was using multiple photos, but using a single one makes me use ImageRepository.saveImage

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
EnesKiper: Same as the one above

app/src/main/java/com/github/meeplemeet/model/images/ImageRepository.kt
EnesKiper: You're right

app/src/main/java/com/github/meeplemeet/model/sessions/SessionRepository.kt
EnesKiper: I removed that function since now the upcoming sessions are just the normal sessions since the past sessions are moved in the firestore db

EnesKiper:
I close this PR because it is not properly scoped

PR #215, opened 2025-11-27
211 bug app crashing when refusing location permission

### LINKED ISSUE #211: Bug: App crashing when refusing location permission
**Overview**  
Opening the map and refusing the location permission causes the app to crash.

**Steps to reproduce**  
1. Launch the app  
2. Open the map view  
3. When prompted for location permission, select "Deny"  
4. The app immediately crashes

**Expected behavior**  
The app should handle the case where location permission is denied gracefully, without crashing.

**Actual behavior**  
The app crashes as soon as the location permission is refused.

**Additional context**  
Crash logs show a `SecurityException` thrown when accessing location APIs without permission.

Atomium90:
**Fix app crashing when refusing to give location permission.**

Related to #211

The crash occurred because `isMyLocationEnabled` was always set to true, which triggered a `SecurityException `when location permission was denied. 
This fix conditions `isMyLocationEnabled `on `permissionGranted`, ensuring the map loads gracefully without crashing.

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
@@ -472,7 +472,8 @@ fun MapScreen(
                       .testTag(MapScreenTestTags.GOOGLE_MAP_SCREEN),
               cameraPositionState = cameraPositionState,
               properties =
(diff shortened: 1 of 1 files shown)


AlexandroHattoum (approved):
Tested the change locally, it seems to work well. I've tried different scenarios and they all seem well handled.
Good job on finding a hidden bug. Nothing else to say from me, once the CI passes you're good to merge.

PR #229, opened 2025-11-27
212 bug map resets to userdefault location when interacting

### LINKED ISSUE #212: Bug: Map resets to user/default location when interacting
**Overview**  
When opening the map, the user is always repositioned either to their current location (if permission is granted) or to the default location (if permission is denied). This happens on every recomposition, which makes navigation frustrating, as the map does not stay where the user moved it.

**Steps to reproduce**  
1. Open the map and grant location permission  
2. The map centers on your current location (expected)  
3. Move the map to another area and open a cluster or preview  
4. The map re-centers back to your location instead of staying where you navigated

**Expected behavior**  
The map should remain at the position where the user navigated, even after recompositions or when opening clusters/previews. Location centering should only happen once at spawn, or when explicitly requested.

**Actual behavior**  
On every recomposition, the map re-centers to the user’s location (or default), overriding the user’s navigation.

Atomium90:
**Fixes map recentering unexpectedly on user/default location**

Related to #212

The issue was caused by two problems:
1. **Camera recenter loop**: the map was re-centered on `userLocation` every time it updated, breaking user interactions
2. **Query restart**: the geo query was re-launched repeatedly instead of only once at startup

**Solution:**
- Added `isCameraCentered` flag to ensure camera centers on `userLocation` only once at startup
- Applied the same logic to the geo query so it starts only once
- Introduced a `DisposableEffect` to keep user location updated in real time without forcing recenter

Tests / validation:
- Map no longer resets when interacting after initial load
- Query runs once and continues to deliver updates
- User location updates smoothly without breaking camera state

No impact on public API – only internal map behavior adjusted.

_Generated by ClaudeAI_

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
@@ -284,6 +284,7 @@ fun MapScreen(
   var userLocation by remember { mutableStateOf<Location?>(null) }
   var isLoadingLocation by remember { mutableStateOf(true) }
   var isCameraCentered by remember { mutableStateOf(false) }
+  var isQueryLaunched by remember { mutableStateOf(false) }
   var includeTypes by remember { mutableStateOf(PinType.entries.toSet()) }
 
   // --- UI controls (filters & creation) ---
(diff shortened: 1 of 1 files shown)


AlexandroHattoum (approved):
Much needed UX improvement, makes the map a lot more usable for users. Once the CI passes and there isn't anything alarming given by SonarCloud, you're good to merge. Thank you for fixing that bug!

PR #235, opened 2025-11-30
Add offline mode support for discussions with message caching

### LINKED ISSUE #221: Implement offline mode for accounts & discussions

AntoineBastide47:
Implements comprehensive offline functionality for discussions, allowing users to view cached discussions and messages when internet connectivity is unavailable.

  Key changes:
  - Enhanced DiscussionViewModel to automatically switch between online/offline data sources
  - Messages are now delayed when sent offline and synced when connection is restored
  
  The offline cache is automatically populated when users view discussions while online, enabling seamless offline access later.


EnesKiper (changes requested):
Great PR, two little problems : 
- When you are offline and go to the discussion screen, you don't see any discussion 
- Secondly, the number of messages sent offline is not capped, which is not an issue on its own, but maybe block sending photos while offline, that can take a lot of memory if not checked

EnesKiper (approved):
Great, you implemented the changes

PR #239, opened 2025-12-01
Prevent creating sessions in the past with inline validation

### LINKED ISSUE #233: Bug: Session creation allowed with past date/time
**Overview**  
It is currently possible to create a session with a date/time in the past.  
The system accepts the input without validation, which leads to inconsistent data (sessions that are already expired at creation => don't visible in overview tab but in history tab).

**Steps to reproduce**  
1. Open the session creation form  
2. Select a date/time that is already in the past  
3. Fill in the other required fields and submit  
4. The session is created successfully, even though the date is invalid
5. Notice that the session does not appear in the **Sessions overview**, but only in the **Sessions historic**

**Expected behavior**  
The app should prevent creating sessions with a past date/time.  
Validation should reject the input and display an error message to the user.

**Actual behavior**  
The session is created normally, with a past date/time, resulting in invalid data.
Thus, the session is immediately placed in the historic list instead of the overview.

AntoineBastide47:
This PR adds validation to prevent users from creating or updating sessions with dates/times in the past. The changes include:

  - Added SelectableDates to the date picker dialog to disable selection of past dates
  - Created isDateTimeInPast() helper function to validate date/time combinations
  - Disabled the "Create" button when selected date/time is in the past
  - Display error messages when users attempt to select past date/times
  - Prevent admins from updating sessions to past date/times on the Session Details screen

  The validation respects the user's timezone and provides clear error messaging to guide users toward selecting valid future dates.

  ---

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/ui/components/SessionComponents.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/components/SessionComponents.kt
@@ -45,6 +45,7 @@ import androidx.compose.material3.IconButtonDefaults
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.MenuAnchorType
 import androidx.compose.material3.RangeSlider
+import androidx.compose.material3.SelectableDates
 import androidx.compose.material3.SliderColors
 import androidx.compose.material3.SliderDefaults
 import androidx.compose.material3.Surface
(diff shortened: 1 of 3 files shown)


Atomium90 (approved):
Thanks for this fix. The error feedback is clearer in this form.  
Once the CI and Sonar checks have passed, you can merge.

PR #241, opened 2025-12-01
Fix session creation using correct repository method

### LINKED ISSUE #240: Bug: Overwrite created session when inviting users that don't accept notifications
Overview
When creating a session, if all the other users in the discussion do not accept invites, they won't be added to the session immediately and the create session button is still visible.

Steps to reproduce
Create a discussion with a user that accepts requests from no one.
Create a session with said user.

Expected behavior
The app should show that the session has been created.

Actual behavior
The app shows the create session icon, allowing the created session to be overriden.

AntoineBastide47:
This PR fixes a bug where session creation was incorrectly using updateSession() instead of createSession(). The changes include:

  - Updated CreateSessionViewModel to call createSession() with the proper method signature
  - Changed participant handling to use varargs format instead of a list
  - Added creator UID as a required parameter for session creation
  - Refined the action icon logic in DiscussionScreen to properly show the "Add Session" icon only for admins when no session exists, and the "Games" icon for users already in a session

  ---

### DIFF
--- a/app/src/androidTest/java/com/github/meeplemeet/integration/FirestoreSessionTests.kt
+++ b/app/src/androidTest/java/com/github/meeplemeet/integration/FirestoreSessionTests.kt
@@ -186,14 +186,7 @@ class FirestoreSessionTests : FirestoreTests() {
   fun canRemoveUserFromSession() = runTest {
     // First create a session with two participants
     viewModel.createSession(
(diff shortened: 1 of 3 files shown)


Atomium90 (approved):
Thanks for those 2 little refactors. LGTM

PR #243, opened 2025-12-01
feat(ui): make messages editable and removable

### LINKED ISSUE #218: Add support for editing and deleting "messages".
- Add a "button/menu" on each message with options to edit or delete (refer to the figma design).
- Editing only available within the allowed time window; if expired, show feedback to the user.
- Comments on posts: only deletable, not editable.

MarcoLourenco55:
This PR introduces the possibility to edit and remove messages from a discussion. It adds the following features :
* A pop-up that appears when pressing a message that we own
* This pop-up allows us to decide to either remove or edit the message

**Visuals**
       src="https://github.com/user-attachments/assets/e40460ef-3d5b-41b0-a8da-2500f2b7cd32">

**References**
* Issue: https://github.com/Meeple-Meet/Meeple-Meet/issues/218

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/ui/discussions/DiscussionScreen.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/discussions/DiscussionScreen.kt
@@ -39,6 +39,7 @@ import androidx.compose.runtime.*
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.blur
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.clipToBounds
 import androidx.compose.ui.geometry.CornerRadius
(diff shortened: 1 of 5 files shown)


dtbcds (changes requested):
I noticed that right now the blur is applied to each individual ChatBubble, which causes a slightly choppy visual experience.

Since you already have a wrapper around each ChatBubble, a better approach would be to repurpose that wrapper instead of blurring the bubble itself:

**1. Use the wrapper to capture bubble coordinates:**

- When a message is selected, the wrapper can use Modifier.onPlaced { coords -> ... } to report the bubble’s bounds.

- This gives you the exact position for any overlay, without modifying the ChatBubble or forcing recompositions.

**2. Draw a single full-screen overlay:**

- Instead of blurring each bubble, draw one Box overlay (similar to the game popup in the ShopScreen.

- This overlay can have a blur and fade effect, keeping everything smooth.

**3. Redraw only the overlay:**

- The wrapper reports the bubble position once, and the overlay handles the visual effect.

- All that remains is redrawing the bubble on top of the box.

dtbcds:
Also consider aligning the edit/delete options with the chat bubbles.

MarcoLourenco55:
Comment https://github.com/Meeple-Meet/Meeple-Meet/pull/243#pullrequestreview-3530696021 solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/243/commits/7220a0b07cbf002f9a6ed933826f244942eedab1

Comment https://github.com/Meeple-Meet/Meeple-Meet/pull/243#issuecomment-3602581451 solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/243/commits/b9bde2d22a7e406995e27e8bc132c2f69245425c

app/src/main/java/com/github/meeplemeet/ui/discussions/DiscussionScreen.kt
```diff
@@ -315,72 +367,241 @@ fun DiscussionScreen(
...
-                            onVote = { optionIndex, isRemoving ->
-                              if (isRemoving) {
-                                viewModel.removeVoteFromPollAsync(
-                                    discussion.uid, message.uid, account, optionIndex)
-                              } else {
-                                viewModel.voteOnPollAsync(
-                                    discussion.uid, message.uid, account, optionIndex)
+                    // Selected message overlay
+                    if (hasSelection) {
+                      val overlayAnchor = anchor!!
```
dtbcds: As sonar mentioned, remove this useless non-null assertion.

app/src/main/java/com/github/meeplemeet/ui/discussions/DiscussionScreen.kt
```diff
@@ -315,72 +367,241 @@ fun DiscussionScreen(
...
+                                    .background(
+                                        MaterialTheme.colorScheme.scrim.copy(
+                                            alpha = UNSELECTED_BACKGROUND_ALPHA))
+                                    .clickable(
+                                        enabled = true,
+                                        indication = null,
+                                        interactionSource =
+                                            remember { MutableInteractionSource() }) {})
+
+                        val message = actionMessage!!
```
dtbcds: Here is another useless non-null assertion you should remove.

app/src/main/java/com/github/meeplemeet/ui/posts/PostScreen.kt
```diff
@@ -259,15 +304,30 @@ fun PostScreen(
...
                 val p = post ?: return@ComposerBar
+                val target = editTarget
+                val editing = target != null
+
                 scope.launch {
                   isSending = true
                   try {
-                    viewModel.addComment(
-                        author = account, post = p, parentId = p.id, text = topComment.trim())
+                    if (editing && target != null) {
```
dtbcds: Here the non-null check always succeeds, you should remove it.

dtbcds (changes requested):
You have checks that seem to be useless and sonar is showing warnings. You should remove these.

app/src/main/java/com/github/meeplemeet/ui/discussions/DiscussionScreen.kt
MarcoLourenco55: The thing was that android studio didn't want me to take out the ```!!``` assertions since it cannot recognize directly (without making compile errors) that this will always pass. To make it understand, I needed to repeat code: replace ```hasSelection``` by it's own same definition.

I prefered to not have repeated code, that's why I kept the ```!!```, but both are fine for me I guess

Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/243/commits/42dd383b4476af8eec97d7774cbd4cdc7d34c6f5

app/src/main/java/com/github/meeplemeet/ui/discussions/DiscussionScreen.kt
MarcoLourenco55: Same comment than the one above

Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/commit/42dd383b4476af8eec97d7774cbd4cdc7d34c6f5

app/src/main/java/com/github/meeplemeet/ui/posts/PostScreen.kt
MarcoLourenco55: In fact, I thought this was for the same reason than the two ones above, that's why I ignored it, but this one is in fact a simple 100% unnecessary test. Thank you for pointing it out !

Solved in commit https://github.com/Meeple-Meet/Meeple-Meet/pull/243/commits/edbf543ee4c8a77453d40e1a4bdd1276dbec83b3

MarcoLourenco55:
Solved in commits https://github.com/Meeple-Meet/Meeple-Meet/pull/243/commits/42dd383b4476af8eec97d7774cbd4cdc7d34c6f5 and https://github.com/Meeple-Meet/Meeple-Meet/pull/243/commits/edbf543ee4c8a77453d40e1a4bdd1276dbec83b3

dtbcds (approved):
Looks good to me! Ready to be merged.

PR #245, opened 2025-12-02
Feat/refactor session creation screen

### LINKED ISSUE #225: Implement redesign of the session creation screen.
- Refactor the session creation screen according to the figma design.

EnesKiper:
**Summary :**
This pull request introduces significant improvements to the session creation UI and underlying logic, focusing on enhancing location and date/time selection, as well as adding handle-based account search functionality. The changes modernize input components, and add new features to the session creation workflow.

**Description :**
* I changed the UI for date and time choosing putting the buttons side by side
* I changed the behavior to add a location to the session, now the user needs to press on the 'Select a location' button to open a pop up and search for the location
* I changed the UI of the 'Participants' session removing the slider, , and changed how the users in the session are displayed

image
image
image

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/ui/components/SessionComponents.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/components/SessionComponents.kt
@@ -16,6 +16,7 @@ import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
@@ -27,9 +28,12 @@ import androidx.compose.material.icons.automirrored.filled.ArrowBack
 import androidx.compose.material.icons.filled.Add
 import androidx.compose.material.icons.filled.CalendarToday
 import androidx.compose.material.icons.filled.Close
+import androidx.compose.material.icons.filled.LocationOn
 import androidx.compose.material.icons.filled.Search
 import androidx.compose.material.icons.filled.Timer
 import androidx.compose.material3.AlertDialog
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.CenterAlignedTopAppBar
 import androidx.compose.material3.DatePicker
 import androidx.compose.material3.DatePickerDefaults
@@ -51,6 +55,7 @@ import androidx.compose.material3.SliderDefaults
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
+import androidx.compose.material3.TextField
 import androidx.compose.material3.TextFieldDefaults
 import androidx.compose.material3.TimeInput
 import androidx.compose.material3.TimePickerDefaults
@@ -66,7 +71,9 @@ import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.onFocusChanged
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.RectangleShape
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.input.ImeAction
@@ -87,6 +94,9 @@ import com.github.meeplemeet.model.space_renter.SpaceRenterSearchViewModel
 import com.github.meeplemeet.ui.FocusableInputField
 import com.github.meeplemeet.ui.navigation.NavigationTestTags
 import com.github.meeplemeet.ui.sessions.SessionTestTags
+import com.github.meeplemeet.ui.sessions.SessionTestTags.LOCATION_PICKER_BUTTON
+import com.github.meeplemeet.ui.sessions.SessionTestTags.LOCATION_PICKER_DIALOG
+import com.github.meeplemeet.ui.sessions.isDateTimeInPast
 import com.github.meeplemeet.ui.theme.AppColors
 import com.github.meeplemeet.ui.theme.Dimensions
 import java.time.Instant
@@ -121,6 +131,9 @@ object ComponentsTestTags {
 /** Common labels, placeholders, and button texts used across components. */
 private const val LABEL_DATE = "Date"
 private const val LABEL_TIME = "Time"
+private const val LABEL_SELECT_DATE = "Select a date"
+private const val LABEL_SELECT_TIME = "Select a time"
+private const val TEXT_SELECT_LOCATION = "Select a location"
 private const val LABEL_LOCATION = "Location"
 private const val PLACEHOLDER_LOCATION = "Enter an address"
 private const val PLACEHOLDER_SEARCH_GAMES = "Search games"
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/meeplemeet/ui/components/SessionComponents.kt
```diff
@@ -288,6 +301,70 @@ fun IconTextField(
...
+ */
+@Composable
+fun IconTextFieldNew(
+    value: String,
+    onValueChange: (String) -> Unit,
+    placeholder: String,
+    editable: Boolean = true,
+    leadingIcon: @Composable (() -> Unit)? = null,
+    trailingIcon: @Composable (() -> Unit)? = null,
+    textStyle: TextStyle = MaterialTheme.typography.bodySmall,
```
AlexandroHattoum: I notice `textStyle` is always assigned to the same value, it's default value. So there's no reason to pass it as a parameter. That will remove your Sonar warning.

AlexandroHattoum (changes requested):
### Overall
Decent UI changes, there's a couple of points I left throughout the UI files to adjust and fix. Documentation is okay and you've got enough line coverage, nothing to add here.

### Code
- Your code seems to be often bloated by either extra parameters that could be hardcoded or by previous code that still lives within your composables. As an example, `UserChipsGrid` still has the old "+ Add participants" button, even though this has been discarded from both create and edit session Figmas. 

- Both screens share intertwined code, which is weird since there's already a file for shared session components. I know this is not strictly part of your task but jumping between 3 files to understand what's happening in one is just a burden for bugfixes and bad for traceability.
- You should add a limit to how many characters can one input for a title, else this could be abused.

### Other
You've got 2 sonar issues that are in principle easy to fix, I'll let you take a look at them.
There's a weird recomposition issue when I create the session, it could be my emulator, but here's a video for investigation.

https://github.com/user-attachments/assets/fb0079c7-2566-4069-9ec1-d6c8eca0e591

app/src/main/java/com/github/meeplemeet/ui/components/SessionComponents.kt
```diff
@@ -755,21 +827,146 @@ fun TimePickerField(
...
+ * A combined date and time picker component.
+ *
+ * @param date The currently selected date.
+ * @param time The currently selected time.
+ * @param onDateChange Callback function to be invoked when a new date is selected.
+ * @param onFocusChanged Callback function to be invoked when the focus state changes.
+ * @param onTimeChange Callback function to be invoked when a new time is selected.
+ */
+@Composable
+fun DateAndTimePicker(
```
AlexandroHattoum, line 851: This composable does not match the figma on multiple standpoints:
- Icons are not colored in AppColors.neutral
- Icons should not be the only clickable item. Rather the whole line should be clickable, this is a lot more mobile friendly
- Add the chevron icon that's exists in the Figma, it is a visual cue to the user that this is clickable.
- Round the corners of the entire row, it's pretty blocky now.

Reference to Figma design:

Image

app/src/main/java/com/github/meeplemeet/ui/components/SessionComponents.kt
```diff
@@ -755,21 +827,146 @@ fun TimePickerField(
...
+ * A button that opens a dialog to search and select a session location.
+ *
+ * @param account The user's account.
+ * @param discussion The discussion associated with the session.
+ * @param viewModel The CreateSessionViewModel for managing session state.
+ * @param buttonTestTag Test tag for the location picker button.
+ * @param dialogTestTag Test tag for the location picker dialog.
+ */
+@Composable
+fun SessionLocationSearchButton(
```
AlexandroHattoum, line 897: To match the Figma design:
- Round the corners of the row.
- Color the icon with AppColors.neutral

This is another piece of feedback, but could you try to make the popup a bit wider? Because it's hard to read a location that spans on 6 lines, even with a reduction of 1 line it would be much cleaner.

Reference to Figma design:

Image

app/src/main/java/com/github/meeplemeet/ui/sessions/CreateSessionScreen.kt
AlexandroHattoum, line 479: Consider reducing the amount of parameters this composable takes. There's a lot of parameters that are probably single use and could even be hardcoded directly. 
I know it's not a nice task but your composable currently has 16 parameters.

app/src/main/java/com/github/meeplemeet/ui/sessions/CreateSessionScreen.kt
```diff
@@ -511,70 +547,45 @@ fun ParticipantsSection(
...
-                                CircleShape)
-                            .padding(
-                                horizontal = Dimensions.Padding.extraMedium,
-                                vertical = CountBubbleVerticalPadding))
-              }
-        }
+        Spacer(Modifier.height(Dimensions.Spacing.medium))
+
+        // Member search field
+        MemberSearchField(
```
AlexandroHattoum: Search should only be for discussion members, I noticed I could add a member that was not in the discussion.

app/src/main/java/com/github/meeplemeet/ui/sessions/SessionDetailsScreen.kt
```diff
@@ -751,43 +738,64 @@ fun UserChip(
...
       modifier = modifier,
-      colors =
-          InputChipDefaults.inputChipColors(
-              labelColor = AppColors.textIcons,
-          ),
-      shape = appShapes.extraLarge)
+      shape = appShapes.medium,
+      color = AppColors.secondary,
+      tonalElevation = Dimensions.Elevation.minimal,
+      border = BorderStroke(Dimensions.DividerThickness.thin, AppColors.divider)) {
```
AlexandroHattoum: I like the idea of adding the handle and the username, it's easier to know who's who.
But the UI card looks a bit odd, I removed the border, removed some padding and changed the background color to blend in with the screen, I think it looks better now. The border could be discussed if it needs to be added back in, but here's how I got it to look with a few minor modifications:

Image

AlexandroHattoum (changes requested):
Thank you for the UI changes, the screen is a lot more fluid and clean now. I see you've also trimmed some composable declarations to take less parameters, that's great.

2-3 details to further enhance the UI and provide better UX for our users.
- Section headers should be bigger as that creates visual separation for the user. Right now they're the same size/smaller than the text in the screen. Section headers = "Basic Info", "Where & When" and "Participants".
- You should add back the search bar to add members from a discussion. Simply limit it to users in the discussion. 
Here's a very likely scenario: suppose you removed a user from the session participants per accident (finger slipped, removed the wrong user due to similar names, conversation misunderstandings...) you'd have to restart everything in the creation process to add this user back. Going to the edit screen to add back the missing participant is also not a good user flow.

app/src/main/java/com/github/meeplemeet/ui/sessions/SessionDetailsScreen.kt
```diff
@@ -749,45 +681,77 @@ fun UserChip(
...
+              Row(
+                  verticalAlignment = Alignment.CenterVertically,
+                  horizontalArrangement = Arrangement.spacedBy(Dimensions.Spacing.large),
+                  modifier = Modifier.weight(1f)) {
+                    // Avatar
+                    Box(
+                        modifier =
+                            Modifier.size(Dimensions.AvatarSize.small)
+                                .clip(CircleShape)
+                                .background(Color.LightGray),
```
AlexandroHattoum: Instead of `Color.LightGray` use `AppColors.divider`.

AlexandroHattoum (changes requested):
A color to change (left a comment for the exact line in SessionDetailsScreen) and a bug to fix, and then it's all good.
Bug: The person who created the session cannot remove himself through the checkboxes. I would say to just remove the checkbox for the session creator in the participants section.

AlexandroHattoum (approved):
All good for me now. I'd recommend you check out your SonarCloud issues though and either accept/resolve them.
Once the CI passes, you're good to merge.

PR #250, opened 2025-12-02
Feat/offline posts

### LINKED ISSUE #222: Implement offline mode for posts

EnesKiper:
**Summary :**
This PR adds caching, and handling general behavior related to posts when offline.

**Description :**
* I added fields in the offline mode object to store cached posts and a queue to cache the posts created while offline
* I added the functions needed to cache posts, fetch the posts from the cache or the db and synchronize the db when back online
* I added backend logic in the offline mode manager to call needed function while online and offline
* I delegated logic from the view models to the offline mode manager


EnesKiper:
The branch was based on main, but it should be based on offline-mode

PR #251, opened 2025-12-02
Feat/offline posts

### LINKED ISSUE #222: Implement offline mode for posts

EnesKiper:
**Summary :**
This PR adds caching, and handling general behavior related to posts when offline.

**Description :**
* Added fields in the offline mode object to store cached posts and a queue to cache the posts created while offline
* Added the functions needed to cache posts, fetch the posts from the cache or the db and synchronize the db when back online
* Added backend logic in the offline mode manager to call needed function while online and offline
* Delegated logic from the view models to the offline mode manager


AntoineBastide47 (changes requested):
Overall a very solid PR.

The architecture is good, but there are significant performance issues with:
  1. Sequential network calls (should be parallel)
  2. Excessive object copying in comment trees
  3. Redundant sorting operations

  For typical usage (3 cached posts, <50 comments per post), it will work fine. But with deeper comment threads or slower networks, users will notice lag. 

Now we need to wait for CI results

AntoineBastide47 (approved):

PR #258, opened 2025-12-03
Introduce Cloud Functions integration with BGG API, caching, and test coverage

### LINKED ISSUE #259: Migrate BggGameRepository to Cloud Functions
### Overview
- Move the existing `BggGameRepository` logic into Firebase Cloud Functions.
- Expose endpoints with efficient caching and coalescing if possible.
- Ensure the repository calls are optimized to reduce redundant API requests.

Atomium90:
### Endpoints
- Add `getGamesByIds` with coalesced batching (up to 20 IDs per batch)
- Add `searchGames` with ranking (exact match → substring → Levenshtein) and `maxResults` cap
- Add `ping` endpoint for basic health checks

### Caching & Coalescing
- Add L1 in‑memory cache with TTL and simple eviction
- Add L2 Firestore cache storing `value` + `updatedAt` (using `Timestamp.now()`)
- Implement debounced request coalescing for `getGamesByIds`

### Parsing
- Parse BGG XML into game output, including optional fields: recommended players, genres, playtime, and minAge
- Handle multiple `<link>` entries to build genres from `boardgamecategory`

### CI & Configuration
- Configure Firebase emulators and CI workflow for functions endpoint.
- Set global function options (`maxInstances`)
- Use `BGG_API_TOKEN` as a function secret for authenticated BGG requests
- Add Gradle build config fields for `GAME_API_HOST` (debug vs release)

### Tests
- Add test suite for CloudBggGameRepository
- Add Jest test suite for Cloud Functions (`getGamesByIds`, `searchGames`, `ping`)
- Mock `firebase-admin` (firestore + `Timestamp`) and BGG API with `nock`

### Notes
- Cloud Functions will be deployed in a following PR  
- CloudBggGameRepository will replace FirestoreGameRepository in a following PR
- `package-lock.json` is a default config file, I didn't add 10k LOC here

_Reformulated with Copilot._


app/src/main/java/com/github/meeplemeet/model/shared/game/CloudBggGameRepository.kt
```diff
@@ -0,0 +1,186 @@
...
+   * Fetches multiple games by their IDs.
+   *
+   * @param gameIDs One or more game IDs to fetch.
+   * @return A list of [Game] objects.
+   * @throws IllegalArgumentException if more than 20 IDs are provided.
+   * @throws GameFetchException if fetching fails.
+   */
+  override suspend fun getGamesById(vararg gameIDs: String): List<Game> =
+      withContext(ioDispatcher) {
+        require(gameIDs.size <= 20) { "A maximum of 20 IDs can be requested at once." }
```
AntoineBastide47: This should be checked before opening a new thread.

app/src/main/java/com/github/meeplemeet/model/shared/game/CloudBggGameRepository.kt
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Fetches a single game by its unique ID.
+   *
+   * @param gameID The game ID to fetch.
+   * @return The [Game] object corresponding to the given ID.
+   * @throws GameFetchException if fetching fails.
+   */
+  override suspend fun getGameById(gameID: String): Game =
+      withContext(ioDispatcher) {
+        return@withContext getGamesById(gameID).first()
```
AntoineBastide47, line 71: Why aren't you handling the exceptions here ?

app/build.gradle.kts
AntoineBastide47, line 1: Shouldn't this be moved to `local.properties` ?

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+function formatLog(obj: Record<string, any> | undefined) {
+  if (!obj) return {};
+  const out: Record<string, any> = {};
+  for (const k of Object.keys(obj)) {
+    out[k] = _safeSerialize(obj[k]);
+  }
+  return out;
+}
+
+function safeLog(message: string, meta?: Record<string, any>) {
```
AntoineBastide47: Why are these called safe** ? There is no try catch block in them so just call them `logDebug`, `logWarning` and `logError`.

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+}
+function safeError(message: string, meta?: Record<string, any>) {
+  if (meta) logger.error(message, formatLog(meta));
+  else logger.error(message);
+}
+
+// --------------------
+// Cache config
+// --------------------
+const GAME_TTL_MS = 24 * 60 * 60 * 1000;
```
AntoineBastide47, line 88: Why do you use 2 constants for the exact same value ?

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+const SEARCH_TTL_MS = 24 * 60 * 60 * 1000;
+const GAME_CACHE_MAX = 1000;
+const SEARCH_CACHE_MAX = 100;
+const GAME_CACHE_COLLECTION = "gameCache";
+const SEARCH_CACHE_COLLECTION = "searchCache";
+
+// Simple L1 in-memory caches (non-persistent)
+type CacheEntry<T> = { value: T; expireAt: number };
+const gameCache = new Map<string, CacheEntry<any>>();
```
AntoineBastide47: Wouldn't it be better to have a simple cache class with get and set so you don't have to pass the caches as parameters ?

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+    };
+
+    pendingBatches.push(pb);
+  });
+}
+
+// --------------------
+// BGG Types
+// --------------------
+type GameOut = {
```
AntoineBastide47: Move the types to the top of the file as this helps with understanding.

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+  timer: NodeJS.Timeout;
+};
+
+const pendingBatches: PendingBatch[] = [];
+
+function scheduleBatchFetch(ids: string[]): Promise<Record<string, GameOut | null>> {
+  return new Promise((resolve, reject) => {
+    for (const pb of pendingBatches) {
+      const combined = new Set([...pb.ids, ...ids]);
+      if (combined.size <= 20) {
```
AntoineBastide47: How does this work for this case:
pendingBatches = [3, 3, 3, 3, 3, 3, 1] (only the amount of ids for simplicity)
ids = 1

Shouldn't you combine as many as possible instead of just the first ?

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+  textNodeName: "#text",
+  ignoreDeclaration: true,
+  removeNSPrefix: true,
+  isArray: (tagName) => tagName === "link",
+});
+
+// --------------------
+// Helpers
+// --------------------
+function safeText(node: any): string | null {
```
AntoineBastide47: This should be named `extractText` for clarity reasons

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+ * Query: ids=1,2,3
+ * Behaviour:
+ *  - check L1 cache per id
+ *  - if some ids missing -> check L2 cache
+ *  - if still missing -> single BGG call for missing ids
+ *  - parse returned items, store each parsed game in L1 + L2 cache
+ *  - return ordered array of games (only successfully parsed ones)
+ */
+export const getGamesByIds = onRequest(
+  { secrets: ["BGG_API_TOKEN"] },
```
AntoineBastide47: Shouldn't this be process.env.BGG_API_TOKEN ?

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+              await setToL2Cache(GAME_CACHE_COLLECTION, g.uid, g).catch((e) =>
+                safeWarn("setToL2Cache rejected", { collection: GAME_CACHE_COLLECTION, key: g.uid, err: e })
+              );
+              cachedById[g.uid] = g;
+            } else {
+              safeWarn("coalesced fetch returned no game for id", { id });
+            }
+          }
+
+          const ordered = ids.map((id) => cachedById[id] ?? null).filter(Boolean);
```
AntoineBastide47: Terrible naming scheme as there is no ordering here.

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+        } catch (err: any) {
+          safeError("Error in coalesced getGamesByIds", {
+            err: err instanceof Error ? { message: err.message, stack: err.stack } : _safeSerialize(err),
+          });
+          res.status(502).json({ error: "BGG fetch failed", message: err?.message });
+          return;
+        }
+      } else {
+        // all cached
+        const ordered = ids.map((id) => cachedById[id]).filter(Boolean);
```
AntoineBastide47: Terrible naming scheme as there is no ordering here.

cloud-functions/bgg-functions/src/index.ts
```diff
@@ -0,0 +1,644 @@
...
+/**
+ * searchGames
+ * Query: query=..., maxResults=..., ignoreCase=true|false
+ * Behaviour:
+ *  - check L1 cache for query key (query normalized + ignoreCase + maxResults)
+ *  - if miss -> check L2 cache
+ *  - if still miss -> fetch BGG search, parse, rank, store in caches
+ */
+export const searchGames = onRequest(
+  { secrets: ["BGG_API_TOKEN"] },
```
AntoineBastide47: Shouldn't this be process.env.BGG_API_TOKEN

cloud-functions/bgg-functions/src/index.ts
AntoineBastide47, line 1: Add more docs as this is typescript code which is harder to understand. And more docs to clearly explain everything the functions do.

AntoineBastide47 (changes requested):
A good addition though some changes still need to be made.

app/build.gradle.kts
Atomium90: Not needed. It's the cloud functions that need to be secure. The URL can be exposed.

app/src/main/java/com/github/meeplemeet/model/shared/game/CloudBggGameRepository.kt
Atomium90: Exceptions are handled by getGamesById and propagated if necessary; no additional try/catch needed here.

app/src/main/java/com/github/meeplemeet/model/shared/game/CloudBggGameRepository.kt
Atomium90: will be.

cloud-functions/bgg-functions/src/index.ts
Atomium90: fair point.

cloud-functions/bgg-functions/src/index.ts
Atomium90: It’s “safe” in the sense that it avoids runtime issues when logging structured objects; not about try/catch. Could rename, but current naming is consistent.

cloud-functions/bgg-functions/src/index.ts
Atomium90: for maintainability

cloud-functions/bgg-functions/src/index.ts
Atomium90: Not really, you would need to handle overlapping keys and TTLs explicitly, so keeping it simple avoids that complexity.

cloud-functions/bgg-functions/src/index.ts
Atomium90: will do

cloud-functions/bgg-functions/src/index.ts
Atomium90: could do

cloud-functions/bgg-functions/src/index.ts
Atomium90: Its how it work with the new Cloud functins API

cloud-functions/bgg-functions/src/index.ts
Atomium90: Its how it work with the new Cloud functins API

cloud-functions/bgg-functions/src/index.ts
Atomium90: will do

cloud-functions/bgg-functions/src/index.ts
Atomium90: will do

cloud-functions/bgg-functions/src/index.ts
Atomium90: Yes, currently we just append to the first batch that can take the new IDs. 
This can be optimized later on

cloud-functions/bgg-functions/src/index.ts
AntoineBastide47: The current naming is missleading so please rename them

cloud-functions/bgg-functions/src/index.ts
AntoineBastide47: The cache and TTL would live in the class and you would call class.get<T>(key) or class.set<T>(key, value), creating a simpler api. The class constructor would take in T and TTL.

cloud-functions/bgg-functions/src/index.ts
AntoineBastide47: You can and should optimize it now, it isn't that hard, just more combined out of the loop and check if combined + ids >= 20 then send and return.

AntoineBastide47 (approved):

PR #279, opened 2025-12-04
BugFix: Notifications Randomly Delete

### LINKED ISSUE #278: Notifications Randomly Delete
### Overview
Bug with handling notifications upon fetching, perfectly valid notifications (no data is malformed, no VM/repo errors) are deleted.

### Steps to reproduce
A bit random, but the idea is that there may be other (non-harmful) exceptions that get thrown upon fetching. The old notification VM simply used a try-catch clause to catch all exceptions, assuming that there would only be an exception when the repo threw due to an unexisting account/discussion.

### Expected behavior
Only notifications from a deleted account/discussion/session should be automatically deleted.

### Actual behavior
Perfectly valid notifications would be deleted upon loading the notifications screen.

AlexandroHattoum:
This PR fixes a bug with the `NotificationsTab.kt` screen. Previously to handle notifications that came from a deleted object (session/discussion/account) I had simply used a try-catch clause, thinking that the only time where an exception could be thrown would be if the data fetched from the repo was invalid.
It turns out there may be other times where exceptions are thrown (but do not cause a crash in the app at all), think of data taking too long to load for example. What this caused in practice was the deletion of perfectly correct notifications.

The fix was simple, only catch the actual exception that I have to handle (AccountNotExist and DiscussionNotExist).

### DIFF
--- a/app/src/androidTest/java/com/github/meeplemeet/integration/NotificationsViewModelTest.kt
+++ b/app/src/androidTest/java/com/github/meeplemeet/integration/NotificationsViewModelTest.kt
@@ -5,13 +5,13 @@ import com.github.meeplemeet.model.account.Account
 import com.github.meeplemeet.model.account.NotificationsViewModel
 import com.github.meeplemeet.model.discussions.Discussion
 import com.github.meeplemeet.utils.FirestoreTests
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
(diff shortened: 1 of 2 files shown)


Atomium90 (approved):
That's a good fix for now. As mentioned, we can change the auto-delete scope later on. Your fix is well tested. 
That's non blocking, but next time try to use the object with an import, instead of using the full path (com.github.meeplemeet.model.someException).

PR #291, opened 2025-12-09
Move NavigationVM logic to MainActivityVM

### LINKED ISSUE #290: Move NavigationVM logic to MainActivityVM
Each screen should only have 1 viewModel. Currently `MainActivity.kt` uses 2 viewModels. The goal is to centralize this logic into MainActivityViewModel and remove NavigationViewModel.

AlexandroHattoum:
This PR simply centralizes the logic of `NavigationViewmodel` and `LocalNavigationVM` into `MainActivityViewmodel`. Previously `NavigationVM` was used to get a notifications count in the bottom bar on the profile icon and it was instantiated through `MainActivity.kt`, which already has it's own VM.

Since each screen should only have 1 Viewmodel, this PR moves this logic to `MainActivityViewModel`, making it such that each screen that uses `BottomNavigationMenu` needs to take an `unreadCount` parameter provided by MainActivity.kt.

These changes keep the convention of 1 viewModel per screen respected without having to instantiate the viewModel in twice.


app/src/main/java/com/github/meeplemeet/MainActivity.kt
```diff
@@ -188,6 +186,7 @@ class MainActivity : ComponentActivity() {
   }
 }
 
+@SuppressLint("StateFlowValueCalledInComposition")
```
AntoineBastide47: Remove this

app/src/main/java/com/github/meeplemeet/MainActivity.kt
```diff
@@ -305,6 +300,7 @@ fun MeepleMeetApp(
                   discussionId = it.uid
                   navigationActions.navigateTo(MeepleMeetScreen.Discussion)
                 },
+                unreadCount = viewModel.unreadCount.value
```
AntoineBastide47: You can't do this, you need to do it like so:
```kotlin
val unreadCount by viewModel.unreadCount.collectAsState()
```
And pass that to the screens instead of `viewModel.unreadCount.value`.

AntoineBastide47 (changes requested):

app/src/main/java/com/github/meeplemeet/MainActivity.kt
AlexandroHattoum: I fixed this in my latest commit, thank you for pointing it out.

app/src/main/java/com/github/meeplemeet/MainActivity.kt
AlexandroHattoum: Done.

AntoineBastide47 (approved):

PR #292, opened 2025-12-09
CI tests optimization

### LINKED ISSUE #272: Fix the CI environnements and randomly failing tests
Currently, the CI has a big problem where some tests will fail randomly in the CI's android's environnement whereas 99% of those never fail for anyone else locally. This task's goal is to investigate this problem in environnement and fix it. Or find workarounds to make the tests work.

dtbcds:
## Overview

This PR improves the reliability of the test suite by removing nondeterministic timing behavior that caused CI to fail randomly.
 What’s Changed

Replaced usages of `Thread.sleep` in tests with deterministic alternatives.

Updated affected tests to rely on explicit signals/conditions instead of timing assumptions.


AntoineBastide47 (approved):
LGTM

PR #295, opened 2025-12-10
Filter notifications from blocked users

### LINKED ISSUE #266: Delete notifications from blocked users
Currently, accounts will still receive notifications from blocked users.

The goal of this task is to prevent that. To do so, one possible solution would be to add a LaunchedEffect(account.notifications) in `MainActivity` to filter them.

AntoineBastide47:
Notifications sent by blocked users are now automatically filtered out when the account data is loaded. This prevents users from seeing friend requests or discussion invites from accounts they have blocked.

  The notification data model has been refactored to separate senderId and discussionId fields (previously combined as senderOrDiscussionId), allowing more precise filtering and clearer semantics for different
  notification types.

  ---


app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
```diff
@@ -35,23 +32,7 @@ const val TAG = "MainActivityViewModel"
...
-    private val shopRepository: ShopRepository = RepositoryProvider.shops,
 ) : AuthenticationViewModel() {
-
-  /**
-   * Signs out the current user and clears all cached flows.
-   *
-   * Overrides the parent signOut method to ensure that all Firestore listeners are removed when the
-   * user signs out.
-   */
-  override fun signOut() {
```
Atomium90, line 48: Maybe some insight on why this function was removed ?

Atomium90 (changes requested):
Overall that's a good work.

Splitting `senderId` and `discussionId `improves maintainability and testability, even if only one is used at a time.
The blocking flow works correctly with session and discussion.
However, when a user is blocked via search, (they appear in the blocklist), if this user send a friend request :

- The request is still received.
- The user is then removed from the blocklist.

Please fix this behavior so that blocking remains consistent and prevents unwanted interactions.

After this fix, I will redo the review to either approve or request changes depending on Sonar analysis.

_Reformulated with Copilot_

app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
AntoineBastide47: It wasn't used anywhere, the clean up commit removed unused code.

Atomium90 (commented):
Thanks for the changes, the friend request with a blocked user now works correctly.
I wil comment my review until the sonar report is on, and at that point I will either approve, or request changes.

Atomium90 (approved):
Sonar is good, you cn merge.

PR #296, opened 2025-12-10
Feat/firebaseauth email change

### LINKED ISSUE #262: Prevent email duplications when changing emails
Currently, when the user changes their email, there is no double check with the firebase authentification module to make sure the new email is not already in use.

The goal of this task is to prevent that. The workflow when a new email is inputed would be:
1) Validate the format of the email
2) Validate with firebase authentication that the email is not already in use
3) Change the email in firebase authentification

As a bonus, please also remove the email field from the `Account` data structure as it can be accessed using `FirebaseProvider.auth.currentUser?.email`.

EnesKiper:
**Summary :**
This PR handles email changing on the firebase authentication side and adds a synchronnization for the firestore database email field of the account. The email change is made once the new email address is verified. 

**Description :**
* I added fields in the EmailSettings screen to be able to change the email by having an email field, a field to confirm the new email, and a field to enter the password needed for reauthentication to make firebase changes
* I added a function to check the new email is not already used
* I added the function to change the email, this function uses the verifyEmailBeforeUpdate, that sends a verification email to the new email address, and the email address is changed in firebase authentication when the link is clicked. The user is automatically logged out from the app when it is done. 
* I added email synchronization for the firestore db when the user logs in to the app

Note for the reviewer : There is a bug that makes the app crash from time to time on log out, but you can check firebase to see that the changes are correctly applied


app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
```diff
@@ -378,6 +379,25 @@ class AccountRepository :
...
   }
 
+  /**
+   * Checks if an email is already in use by another account.
+   *
+   * @param email The email to check
+   * @param currentUserId The current user's ID to exclude from the check
+   * @return true if the email is in use by another account, false otherwise
+   */
+  suspend fun isEmailInUse(email: String, currentUserId: String): Boolean {
```
AntoineBastide47, line 389: You should use the firebase authentication to check this instead and look for an `email already in use` error code.

app/src/main/java/com/github/meeplemeet/model/auth/AuthenticationRepository.kt
```diff
@@ -273,6 +297,111 @@ class AuthenticationRepository(
...
+      } catch (_: Exception) {
+        // Reauthentication failed - likely wrong password
+        return Result.failure(IllegalStateException(WRONG_PASSWORD_MSG))
+      }
+
+      // Step 5: Send verification email and schedule email update
+      // Firebase requires verification before changing email for security
+      // The email in Firebase Auth will update after user clicks the verification link
+      try {
+        currentUser.verifyBeforeUpdateEmail(newEmail).await()
```
AntoineBastide47, line 358: Instead of using the db to check if the email is already in use, use firebase authentication like so (might not work out of the box, this is gpt code, so there might be some errors):
```kotlin
try {
        Firebase.auth.currentUser!!
            .verifyBeforeUpdateEmail(newEmail)
            .await()
    } catch (e: Exception) {
        if (e is FirebaseAuthUserCollisionException &&
            e.errorCode == "ERROR_EMAIL_ALREADY_IN_USE"
        ) {
            // email already bound to another account
        }
        throw e
    }
```

app/src/main/java/com/github/meeplemeet/model/auth/AuthenticationRepository.kt
```diff
@@ -273,6 +297,111 @@ class AuthenticationRepository(
...
+   * NOTE ON DUPLICATE EMAIL DETECTION:
+   * - This method checks Firestore for duplicate emails before sending the verification email
+   * - This prevents the common case where another account in your system uses the email
+   * - Firebase Auth will also check when the verification link is clicked
+   *
+   * @param newEmail The new email address to set for the user
+   * @param currentPassword The user's current password for reauthentication
+   * @return Result indicating success, or failure with a user-friendly error message
+   */
+  suspend fun updateEmail(newEmail: String, currentPassword: String): Result<Unit> {
```
AntoineBastide47, line 325: Is it normal that both emails get sent the verification link ? And that I wasn't signed out after clicking the link (as suggested by the confimation message)

AntoineBastide47 (changes requested):
Very good PR, a small change is required though. Also you need to fix your CI issues.

app/src/main/java/com/github/meeplemeet/model/auth/AuthenticationRepository.kt
EnesKiper: I checked online and by testing the app, and the function to update the email does not seem to throw that error even though in the backend the verification email is not.
I still need to show the user that the email is a duplicate, so i will keep it like that.

app/src/main/java/com/github/meeplemeet/model/account/AccountRepository.kt
EnesKiper: Same as below

app/src/main/java/com/github/meeplemeet/model/auth/AuthenticationRepository.kt
EnesKiper: Can you check again, because i don't have that problem, for me a warning email is sent to warn that the email has changed.
The update is made when going to the profile screen or starting again the app.
I changed the confirmation message to ask the user to log back in with their new email.

app/src/main/java/com/github/meeplemeet/model/auth/AuthenticationRepository.kt
AntoineBastide47: Okay

app/src/main/java/com/github/meeplemeet/model/auth/AuthenticationRepository.kt
AntoineBastide47: All good, I miss read the email, my bad. Though I think you should either sign out the user once the button is clicked or clear the form and sync the email field to the new one sent by firebase auth.

AntoineBastide47 (approved):

PR #341, opened 2025-12-12
Goated CI Upgrade

AntoineBastide47:
This PR changes the CI to run the android tests 10 times in parallel and needs at most 1 of the 10 to succeed.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -333,7 +333,7 @@ jobs:
           force-avd-creation: true
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none -skin 1080x2400
           disable-animations: true
-          script: ./gradlew connectedAndroidTest --no-parallel --build-cache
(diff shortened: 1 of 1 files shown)


Atomium90 (changes requested):
Ok that's a very very great improvement. I do see a few things that could be refined (if possible):
- Could you stop a run as soon as one test fails? Since that run will be inconclusive, stopping early would save resources.
- As I’m writing this review, two runs have already succeeded, which means the whole CI will pass. However, the other runs are still ongoing. Could you stop them and move directly to the final task once one run succeeds?
- Lastly, maybe limit to 5 runs in parallel for now. If that turns out not to be enough, we can increase later.

AntoineBastide47:
I will leave 10 tasks and add the early fail. The other one isn't possible (according to claude, I can't be bothered to look further into it).

Atomium90:
Fair enough, the early stopping will already be a good optimization.

Atomium90 (approved):
As this is critical, I will accept it now, but it could be intersting at some point to try to optimize this new matrix run to save some resources.

PR #343, opened 2025-12-14
Feat/account info popup

### LINKED ISSUE #315: Implement the account popup when clicking on a profile picture
When pressing on the profile picture, there should be a popup for the account according to the figma.

AlexandroHattoum:
## Description
This PR introduces the `UserProfilePopup`, a popup dialog that enables users to view other's profiles and manage relationships. It is currently wired to `DiscussionScreen.kt` and `PostScreen.kt` and can be easily wired into other screens.

### Key Changes

- **Profile Display:** Detailed view showing the user’s avatar, display name, handle (`@handle`), and description.
- **Offline Mode Support:** Disables the social actions. 
- **Social Actions**:
  - Send Friend Request: Initiates a friend request and sends a notification to the user in question.
  - Block: Blocks the target user which causes their messages to be hidden.
  - Remove Friend: Appears only when both users are already friends, allowing removal.
  All of these actions will show a feedback snackbar for about 2 seconds.

Demo: 

https://github.com/user-attachments/assets/fc431ffc-3e40-4556-bb9b-e4cceba7c1bf

### DIFF
--- a/app/src/androidTest/java/com/github/meeplemeet/ui/DiscussionScreenIntegrationTest.kt
+++ b/app/src/androidTest/java/com/github/meeplemeet/ui/DiscussionScreenIntegrationTest.kt
@@ -55,13 +55,18 @@ class DiscussionScreenIntegrationTest : FirestoreTests() {
             photoUrl = null)
 
     // Create a test discussion    // Create initial account with a long description
-    val longDescription = "Start " + "A very long description that should definitely overflow because it repeats many times. ".repeat(10) + " End"
-    otherUser = accountRepository.createAccount(
+    val longDescription =
+        "Start " +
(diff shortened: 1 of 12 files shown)


Atomium90 (changes requested):
Overall, your addon looks great. Nice idea to use an interface for the popup action.

I have just two minor suggestions:
- Consider adding the popup in the friends screen, so users can see more details about the people they add or request.
- Instead of only showing that a request has been sent, you could update the button to allow canceling the friend request—similar to how it works in the friends screen.

With these small adjustments, the PR is good to merge.

Atomium90 (approved):
Thanks for those changes. it's good to merge!

PR #347, opened 2025-12-14
Feat/move email verification

### LINKED ISSUE #316: Move the email and verification to main tab
Move the email and verification to main tab similar to the figma design.

AlexandroHattoum:
### Description
This PR moves the email verification logic out of the subscaffold "Manage your email" and places it into the main page of the profile screen. The section is only displayed while the user has not verified his email, it is placed within the center of the screen with an eye-catching color to ensure that this step is performed asap by the user.
Email verification is a small step that however unlocks many extra features within the app and a layer of security.

Within this PR there is also a bit of code cleanup within the email related sections of the screen.

Demo:

https://github.com/user-attachments/assets/59ccc435-2da2-4de2-bf5a-1ed354ac7b1b

### DIFF
--- a/app/src/androidTest/java/com/github/meeplemeet/ui/EmailChangeUITest.kt
+++ b/app/src/androidTest/java/com/github/meeplemeet/ui/EmailChangeUITest.kt
@@ -41,10 +41,8 @@ class EmailChangeUITest : FirestoreTests() {
     compose.setContent {
       EmailSection(
           email = "test@example.com",
-          isVerified = true,
           onEmailChange = {},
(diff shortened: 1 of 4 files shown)


AntoineBastide47 (changes requested):
Very good changes, though some smaller ones are still needed:
1) What is this black bare at the top of the screen ? Can you remove it
Image
2) Use the default colours after the email is verified on the input field
3) I can't scroll down to the bottom so the delete button is hidden

AntoineBastide47 (approved):

PR #353, opened 2025-12-16
Fix redundant close button in MarkerPreviewSheet

### LINKED ISSUE #336: Remove the X in pin previews
Image

Atomium90:
The issue was that `MarkerPreviewSheet` exposed both:
1. **Explicit close button** (`PREVIEW_CLOSE_BUTTON`) inside the sheet  
2. **Bottom sheet dismiss** via `onDismissRequest`

This created duplicated logic and inconsistent UX compared to `ClusterPreview`.

### Solution
- Remove the explicit close button from `MarkerPreviewSheet`
- Unify closing behavior to rely solely on `ModalBottomSheet`’s `onDismissRequest`
- Ensure consistency with `ClusterPreview` which already uses the same pattern

### Tests / validation
- Bottom sheet can be dismissed via swipe down or backdrop tap
- `viewModel.clearSelectedPin()` is correctly invoked on dismiss
- No regressions in marker selection or preview display
- Update tests to remove references to `PREVIEW_CLOSE_BUTTON`

_No impact on public API – only internal UI behavior adjusted for consistency._

_Generated with Copilot._

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
+++ b/app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
@@ -463,7 +463,9 @@ fun MapScreen(
         when {
           markerPreview != null -> {
             ModalBottomSheet(
-                sheetState = sheetState, onDismissRequest = { viewModel.clearSelectedPin() }) {
+                sheetState = sheetState,
+                onDismissRequest = { viewModel.clearSelectedPin() },
+                containerColor = AppColors.primary) {
                   if (isLoading) {
                     MarkerPreviewLoadingSheet(pin = uiState.selectedPin)
                   } else {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
AlexandroHattoum, line 465: Could you add here the following line to fix the background color of the popup?
```containerColor = AppColors.primary,```

AlexandroHattoum: Same feedback here:
```containerColor = AppColors.primary,```

AlexandroHattoum (changes requested):
Nothing to say about the removal of the icon, it was done well. Could you however change the modal bottom sheet container color to one of the colors of the app?

AlexandroHattoum (approved):
LGTM!

PR #354, opened 2025-12-16
Fix concurrent modifications linked to posts

AntoineBastide47:
Simple PR to fix concurrent modifications caused by editing posts


AlexandroHattoum (approved):
I've tested it locally, creating posts online and offline no longer crashes the app. LGTM!

PR #357, opened 2025-12-16
Remove firebase storage error logs

### LINKED ISSUE #328: Logcat shows lots of FirebaseStorage errors, fix those
There is a lot of fails in the Logcat, probably more than you can fix in the allocated time frame. So choose wisely the ones you can fix in 4h

AntoineBastide47:
This PR fixes the storage error logs that could be seen in logcat by making sure the image we are fetching actually has an existing stored url before fetching it.


dtbcds (approved):
LGTM

PR #358, opened 2025-12-16
Fix/sessions overview screen stale cache

### LINKED ISSUE #330: In sessions overview, if you click on a first session, and then on a second one, you are redirected to the discussion instead of the session.

EnesKiper:
**Summary :**
This PR solves the bug where clicing on a new session would make the app navigate to the discussion instead of the session.

**Description :**
* I added a timer to wait for the fresh data from the db before loading the data from the cache that might be stale


AntoineBastide47 (changes requested):
LGTM, just waiting for the CI

PR #359, opened 2025-12-16
Add ownership filter and refactor geo-pin handling

### LINKED ISSUE #326: Show owned pins differently
Add a new filter for "my businesses"

Atomium90:
This PR adds an ownership-based filter to the map and cleans up geo-pin creation/update flows to support it efficiently.

### Map filtering
- Add **“My businesses only”** filter, independent from pin type filters
- Filter applies only to **SHOP** and **SPACE** pins (sessions remain visible)
- Centralize filtering logic in `GeoPinWithLocation.matchesFilters`
- Update `MapViewModel` cluster computation to include ownership filtering

### Data & repositories
- Add nullable `ownerId` to `StorableGeoPin`
- Make low-level `upsertGeoPin` private
- Introduce explicit wrappers:
  - Business geo-pin creation (SHOP / SPACE with required `ownerId`)
  - Session geo-pin creation (no ownership)
- Add `updateGeoPinLocation` to update position without overwriting metadata

### UI
- Refactor map filters into a dedicated `FilterPanel` composable
- Add a switch for the ownership filter with clear labeling
- Keep ownership filter independent from pin type chips

### Tests
- Fix geo-pin repository tests impacted by `ownerId` and private upsert
- Add tests for `updateGeoPinLocation`
- Update ViewModel tests for new `getClusters(account)` signature
- Add tests for ownership-based filtering in `MapViewModel`
- Add UI tests for the new filter panel

### Impact
- Enable efficient client-side filtering without extra fetches
- Prevent accidental metadata overwrites when updating locations
- Clearer separation of concerns between creation and update flows

### Testing
- Toggle “My businesses only” and verify only owned shops/spaces remain
- Ensure sessions are always visible
- Update a shop location and confirm ownership data is preserved
- Verify clusters update correctly when filters change

_Reformulated with Copilot._

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/model/map/StorableGeoPin.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/map/StorableGeoPin.kt
@@ -15,8 +15,9 @@ enum class PinType {
  *
  * @property uid Globally unique identifier of the pin (Firestore document ID).
  * @property type Type of the pin (e.g. SHOP, SESSION, SPACE).
+ * @property ownerId Non-null if the pin represent a business (SPACE, SHOP).
  */
-data class StorableGeoPin(val uid: String, val type: PinType)
+data class StorableGeoPin(val uid: String, val type: PinType, val ownerId: String? = null)
 
 /** A geo-pin along with its current geographic location. */
 data class GeoPinWithLocation(val geoPin: StorableGeoPin, val location: GeoPoint)
@@ -26,7 +27,8 @@ data class GeoPinWithLocation(val geoPin: StorableGeoPin, val location: GeoPoint
  *
  * Firestore stores the UID as the document ID, so it is omitted from the stored object.
  */
-@Serializable data class StorableGeoPinNoUid(val type: PinType = PinType.SHOP)
+@Serializable
+data class StorableGeoPinNoUid(val type: PinType = PinType.SHOP, val ownerId: String? = null)
 
 /**
  * Converts a full [StorableGeoPin] into its Firestore-storable form [StorableGeoPinNoUid].
@@ -34,7 +36,7 @@ data class GeoPinWithLocation(val geoPin: StorableGeoPin, val location: GeoPoint
  * @param pin The map pin instance to convert.
  * @return The stripped-down form without UID for storage.
  */
-fun toNoUid(pin: StorableGeoPin): StorableGeoPinNoUid = StorableGeoPinNoUid(pin.type)
+fun toNoUid(pin: StorableGeoPin): StorableGeoPinNoUid = StorableGeoPinNoUid(pin.type, pin.ownerId)
 
 /**
  * Reconstructs a full [StorableGeoPin] object from its Firestore representation.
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/meeplemeet/model/map/StorableGeoPin.kt
```diff
@@ -26,15 +27,16 @@ data class GeoPinWithLocation(val geoPin: StorableGeoPin, val location: GeoPoint
  *
  * Firestore stores the UID as the document ID, so it is omitted from the stored object.
  */
-@Serializable data class StorableGeoPinNoUid(val type: PinType = PinType.SHOP)
+@Serializable
+data class StorableGeoPinNoUid(val type: PinType = PinType.SHOP, val ownerId: String? = null)
```
MarcoLourenco55, line 31: I am not sure I understood why here only ```PinType.SHOP``` is used

app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
```diff
@@ -1517,6 +1473,132 @@ private fun StaticVerticalMapMenu(
...
+                      // Box to center the scaled switch
+                      Box(
+                          modifier = Modifier.size(Dimensions.IconSize.medium),
+                          contentAlignment = Alignment.Center) {
+                            Switch(
+                                checked = showOwnedOnly,
+                                onCheckedChange = { onOwnedOnlyToggle() },
+                                modifier = Modifier.scale(0.7f),
+                                colors =
+                                    SwitchDefaults.colors(
```
MarcoLourenco55, line 1588: (nitpick) Maybe consider making the toggle also AppColors.focus to match the checkboxes colors. But I am not sure this is better. What do you think ?

MarcoLourenco55 (approved):
Great PR, it makes the use of the map way better for business users and it is quite reactive in my phone.

I have only two nitpick questions about your code, but as it is, you can merge !

app/src/main/java/com/github/meeplemeet/model/map/StorableGeoPin.kt
Atomium90: Serializable data class need a default value, so it's for the exact same reason that ownerId is default to null.
I could have default type to SPACE, or SESSION without any impact.

app/src/main/java/com/github/meeplemeet/ui/MapScreen.kt
Atomium90: That's more of a design question. Personally, I think keeping it affirmative (blue) adds a nice visual separation between the checkboxes and the switch (obviously, on top of the clear separation between checkboxes and the switch).
But if you want it focus (yellowy orange)  I can do it.

MarcoLourenco55: I guess that's fine then. I was hesitating myself between the two, so I think keeping it blue is better.

PR #363, opened 2025-12-17
Fix/e2e trial2

AlexandroHattoum:
Fixing stuff


AlexandroHattoum:
This PR was opened as to test things in relation to fixing the End2End tests as they pass locally but fail on the CI. The PR was used as a way to try out new things without having to revert some commits or cause more trouble. As the issues narrowed down and were properly diagnosed, this PR can be closed.

PR #364, opened 2025-12-17
Delete duplicate notifications

AntoineBastide47:
This PR fixes duplicate notifications by deleting all duplicates on load and keeping only the youngest one.

### DIFF
--- a/app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
+++ b/app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
@@ -15,7 +15,6 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.combine
-import kotlinx.coroutines.flow.count
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.launch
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
```diff
@@ -95,21 +83,37 @@ class MainActivityViewModel(
...
                 val filtered =
                     account.notifications.filter {
                       account.relationships[it.senderId] != RelationshipStatus.BLOCKED
                     }
-                val toRemove = account.notifications.filterNot { filtered.contains(it) }
+
+                // Remove duplicates: same fields except uid and sentAt, keep newest
+                val deduplicated =
+                    filtered
+                        .groupBy { n ->
```
MarcoLourenco55, line 95: Just make sure that the fact that the order is not deterministic. If you want a stable ordering for the notifications, maybe add something like ```.sortedByDescending``` lambda.

app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
```diff
@@ -95,21 +83,37 @@ class MainActivityViewModel(
...
+                              n.receiverId,
+                              n.read,
+                              n.type,
+                              n.executed())
+                        }
+                        .values
+                        .map { group -> group.maxBy { it.sentAt } }
+
+                // Delete blocked and duplicate notifications
+                val toRemove = account.notifications.filterNot { deduplicated.contains(it) }
```
MarcoLourenco55, line 108: Isn't it this O(n^2) ? Maybe consider computing a first time a set of all of them and then re-use it, instead of computing this each time.
Something like :
```kotlin
val keep = deduplicated.map { it.uid }.toSet()
val toRemoveUids = account.notifications.map { it.uid }.filter { it !in keep }
```

MarcoLourenco55 (changes requested):
Thank you for this PR. A part from two small specific comments I have nothing to say, the rest LGTM

app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
AntoineBastide47: The sorting is done here: `.map { group -> group.maxBy { it.sentAt } }`

app/src/main/java/com/github/meeplemeet/model/MainActivityViewModel.kt
AntoineBastide47: There will never be more than 25 notifications, so it isn't an issue and you need the whole notification to deduplicate.

MarcoLourenco55 (approved):

AntoineBastide47:
No tests were added because we haven't figured out a way to test state flows. Each time we tried, the tests would always fail.

PR #366, opened 2025-12-17
bug(ui): make session creation in the past error full-line

### LINKED ISSUE #324: The time error message in session create is not on the whole line
Image

MarcoLourenco55:
This PR solves the bug that the error message showned when trying to create and edit a session in the past was only being showned in half part of the screen. It is now full width and centered as you can see in the image below :

image

Ref: issues https://github.com/Meeple-Meet/Meeple-Meet/issues/324


dtbcds (approved):
Much better, LGTM

